PR #1, opened 2015-06-20
Fixed win32 demo

ocornut:
Some fixes I needed to get the win32 demo to compile. Took me a bit of time to figure how to compile file. I eventually used this command-line.
```
cl /I.. win32.c /link gdi32.lib user32.lib
```
Would be good to document that somewhere (or provide a batch file?)


vurtun:
Thanks for your fixes. I personally don't have a windows system so it is always cumbersome to keep it up to date. About the batch file you are probably right. I did not use visual studio directly to compile it since I don't like it. I think I will add batch files as well as make files to build the demos.

PR #11, opened 2015-08-25
Compile on Windows with /Wall /WX

serge-rgb:
- `long *u` to `gui_long *u` in line 508 to suppress an MSVC warning.
- Line 368 MSVC warning. m1 may be uninitialized. 

The part where I might have done something wrong is in lines [4840, 4847]
panel_padding was not being initialized. And item_padding was being initialized as `gui_config_property(config, GUI_PROPERTY_PADDING);` which is not consistent to the rest of the file, where as far as I can tell, item_padding is always initialized with `gui_config_property(config, GUI_PROPERTY_ITEM_PADDING);`

I'm don't understand the code, this was a mechanical substitution, but it seemed safe to assume that using `GUI_PROPERTY_PADDING` for item_padding was a typo. 

In case I'm wrong about item_padding, please ignore the pull request.

Thanks again for making this! Just compiled the opengl.c demo on Linux and it looks great

### DIFF
--- a/gui.c
+++ b/gui.c
@@ -365,7 +365,7 @@ static gui_size
 gui_dtos(char *s, gui_double n)
 {
     gui_int useExp;
-    gui_int digit, m, m1;
+    gui_int digit, m, m1 = 0;
     gui_char *c = s;
     gui_int neg;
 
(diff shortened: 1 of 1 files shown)


vurtun:
Thanks for your pull request. I fixed the `long` to `gui_long` and `m1` bug and will upload the changes
with the next commit. As for the `GUI_PROPERTY_PADDING` it is correct but the variable name was incorrect. I already changed it in my code previously. With a little bit of luck I will have the next update up tomorrow.

PR #25, opened 2015-11-26
windows support

mattn:
fixed all Makefile(s) for windows


vurtun:
Hey and thanks for another great PR.
I looked over it and everything works fine at least on the Linux. I cannot test the
windows version at the moment because my virtual machine just stopped working, so
I will trust your changes on the windows side.

PR #43, opened 2016-01-19
Text input on iOS and tvOS

richi:
This commit does implement text input for iOS and tvOS using the native onscreen keyboard. To show and hide the keyboard have a look at: zr_backend_show_keyboard() / zr_backen_hide_keyboard() at the end of ZahnradBackend.m. To get this to work automatically, we would probably need a call back structure like alloc etc.

### DIFF
--- a/demo/apple/Mac_GL.xcodeproj/project.pbxproj
+++ b/demo/apple/Mac_GL.xcodeproj/project.pbxproj
@@ -27,6 +27,8 @@
 		500B20451C4B43BC0002B53F /* Roboto-Bold.ttf */ = {isa = PBXFileReference; lastKnownFileType = file; path = "Roboto-Bold.ttf"; sourceTree = "<group>"; };
 		500B20461C4B43BC0002B53F /* Roboto-Light.ttf */ = {isa = PBXFileReference; lastKnownFileType = file; path = "Roboto-Light.ttf"; sourceTree = "<group>"; };
 		500B20471C4B43BC0002B53F /* Roboto-Regular.ttf */ = {isa = PBXFileReference; lastKnownFileType = file; path = "Roboto-Regular.ttf"; sourceTree = "<group>"; };
+		504DFDEA1C4EA2CD00D3714A /* zahnrad.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = zahnrad.h; path = ../../../zahnrad.h; sourceTree = "<group>"; };
+		504DFDED1C4EA34800D3714A /* demo.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; name = demo.c; path = ../../demo.c; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
@@ -60,9 +62,11 @@
 		500B201E1C4B41870002B53F /* Mac_GL */ = {
 			isa = PBXGroup;
 			children = (
+				504DFDED1C4EA34800D3714A /* demo.c */,
 				500B203E1C4B43930002B53F /* ZahnradBackend.h */,
 				500B203F1C4B43930002B53F /* ZahnradBackend.m */,
 				500B20231C4B41870002B53F /* main.m */,
+				504DFDEA1C4EA2CD00D3714A /* zahnrad.h */,
 				500B20411C4B43A30002B53F /* zahnrad.c */,
 				500B202A1C4B41880002B53F /* Info.plist */,
 			);
--- a/demo/apple/README.md
+++ b/demo/apple/README.md
@@ -3,7 +3,7 @@ Zahnrad for Mac OS X, iOS and tvOS
 
 This is a bunch of experimental render backends for [Zahnrad](https://github.com/vurtun/zahnrad) a gorgeous immediate mode GUI toolkit.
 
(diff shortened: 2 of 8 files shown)


richi:
Mobile devices usually have an onscreen keyboard and we need a way to activate it, if needed.

For testing, I simply include the following lines

    if (*active)
    {
        void zr_backend_show_keyboard(zr_hash name, struct zr_rect bounds, struct zr_buffer* text);
        zr_backend_show_keyboard(win->edit.name, bounds, buffer);
    }

at the very end of `zr_edit_buffer()`. This was already enough to get it to work, but maybe i did overlook something?

zr_backend_show_keyboard() does call the `showKeyboard:` methods in main.m on iOS and tvOS. There the keyboard is shown as soon as the widget gets activated for the first time. On tvOS things are a little crazy, because one has to build an overlay, that is why I need the rectangle and the current text.

Hiding can be done manually, but opening the keyboard automatically as soon as a widget gets activated would be nice. Could we include this as native functionality through a call back system analog to `struct zr_allocator`. Should be only a very small change. What do you think?

dumblob:
In general we need some way of activating virtual keyboards and similar stuff. I unfortunately won't have time to test it properly (BlackBerry uses QT and I didn't have time to try integrating *zahnrad* into the BB application ecosystem).

Bbut I can surely tell, that each platform will need different data (if any) to be passed through `zr_backend_show_keyboard()` (or similar) and therefore a signature `zr_backend_show_keyboard(zr_hash name, void *payload)` or even `zr_backend_show_keyboard(void *payload)` should be used.

richi:
Sorry, my bad, I was too lazy to type the full prototype for the call back function.

    void zr_backend_show_keyboard(zr_hash name, struct zr_rect bounds, struct zr_buffer* text)

Right now we have the name (hash) of the widget, the bounds and the current text. Of course one should include a payload pointer like in `struct zr_allocator` as well.

That would give something like:

    struct zr_onscreen_keyboard {
        zr_handle userdata;
        void (*show)(zr_handle userdata, zr_hash name, struct zr_rect bounds, struct zr_buffer* text);
        void (*hide)((zr_handle userdata); 
    };

dumblob:
No need to apologize - the overall idea is great.

The proposal `struct zr_onscreen_keyboard` sounds good (maybe we should generalize the name to something like `struct zr_onscreen_non_positional_input`, but shorter). I'm though still a bit afraid of the `struct zr_buffer* text` pointer as the non-positional input can return (and show) other information than just text. Imagine e.g. hand-drawn shape recognition used as input for complex mathematical notation or any other notation which is not defined in utf-8.

On the other hand, usually these special shapes/whatever are being encoded into ASCII anyway, so the pointer to a `zr_buffer` might in theory be enough, but still a bit less convenient than just a plain pointer, which wouldn't encourage one to come up with weird solutions like specific encodings etc.

But I might be totally wrong as my experience with APIs for on-screen input is very limited.

vurtun:
Sound interesting, but before I add anything to the library I want to make sure there is no other way to achieve the same result without touch specific callbacks. As far as I understood the problem is that you have to know if a `zr_edit_string`,  `zr_edit_buffer` and `zr_property_xxx` is activated or deactivated so that you can either activate or deactivate the native onscreen keyboard?
If so both `zr_edit_string` and `zr_edit_buffer` return their current state which could be modified to return a flag to indicate the state and a flag to indicate a state change. Is there a way to use this to create a special version of both on the demo side which checks for a state change and deactivate/activates the onscreen keyboard?
Something like this that could be added to the demo?:
```c
static void
zr_touch_edit_string(struct zr_context *ctx, zr_flags flags, char *buffer, zr_size *len, 
    zr_size max, zr_filter filter)
{
    zr_flags state = zr_edit_string(ctx, flags, buffer, len, max, filter);
    if (state & ZR_EDIT_ACTIVATED)
        zr_backend_show_keyboard(...);
    if (state & ZR_EDIT_DEACTIVATED) 
        zr_backen_hide_keyboard(...);
}
```

richi:
>which could be modified to return a flag to indicate the state and a flag to indicate a state change.

Well, I woul love to push you a little more into the touch direction :) but to be honest, since one has to do a lot of adaptations for a serious touch UI anyhow, your solution will do it just fine. 

With those new flags, defining something like ZR_TOUCH_SCREEN before including demo.c could trigger the implementation and use of `zr_touch_edit_string`. Sound good to me.

vurtun:
No I did not mean to add `zr_touch_edit_string` directly to the library but instead let the user basically create it himself and only provide the tools to make possible.
For example if I add the flags to `zr_edit_xxx` and you call `zr_layout_peek` before calling `zr_edit_xxx` you get both the edit widget state as well as the bounding rectangle, which should provide enough information to make the on screen keyboard work and let the user create his own `zr_touch_edit_xxx` utility functions if he wants to (maybe create a version inside the platform dependent demo code). This on the other hand would free the library from any platform depended code. Or did I misunderstand or do not know about something that would prevent this abstraction?

richi:
> Or did I 

Just a little bit, I did fully agree with you at the first hand. Let's summarize:

* The core will deliver the new activate / deactivate flags (very useful for other things as well BTW).
* Someone (propably me) will implement an example zr_touch_edit... widget in demo.c.
* This example in demo.c will be shielded by a ˋ#if ZR_TOUCH...ˋ statement. 

Right?

vurtun:
Sounds perfect I will rewrite the core and push the changes ASAP.

vurtun:
Changes have been pushed. I tested it and it should work.

dumblob:
I like this solution. Thanks @vurtun.

PR #44, opened 2016-01-20
Show and hide onscreen keyboard

richi:
This contains an example how to automatically show and hide a onscreen keyboard. The logic is implemented in zr_touch_edit_string (ZahnradBackend.m)


vurtun:
I am not quite sure but shouldn't `zr_layout_peek` be called before calling `zr_edit_string` in `zr_touch_edit_string` or does it work nonetheless?

richi:
>  or does it work nonetheless

probably by accident only :)

fixed...

PR #45, opened 2016-02-01
vertexes -> vertices

richi:


vurtun:
As far as I know both `vertexes` and `vertices` are correct plural forms for vertex. But since I absolutely don't care which one is used I will just merge it.

richi:
Thanks! I guess I am somewhat old fashioned :)

PR #60, opened 2016-04-06
Replace an `assert` by `ZR_ASSERT`

adriweb:


vurtun:
Thanks for your pull request. I am currently rewriting some big parts of the library for the upcomming 1.0 release and the bug has already been fixed. I will upload the new version when it is done probably some time this week if everything works out.

PR #61, opened 2016-04-18
Fixes for C++ compatibility, corrected an assert call

angusholder:
There were a few casts from void * missing causing problems when
compiling as C++, also the assert function from <assert.h> was being
used instead of NK_ASSERT(...) in one place.


vurtun:
Thanks for the PR looks good.

PR #62, opened 2016-04-18
Clearing up compiler warnings

angusholder:
Some minor issues: an unused variable, fixed a comparison of int against
float, and a pragma warning (pop) that didn't have a matching pragma
warning (push). Potentially more serious was a possible buffer overrun,
where a pointer was being tested against a null terminator, instead of
the char it points to.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -3113,7 +3113,7 @@ nk_strmatch_fuzzy_text(const char *str, int str_len,
         int rematch = best_letter && nk_to_lower(*best_letter) == nk_to_lower(str_letter);
 
         int advanced = next_match && best_letter;
-        int pattern_repeat = best_letter && pattern_iter != '\0';
+        int pattern_repeat = best_letter && *pattern_iter != '\0';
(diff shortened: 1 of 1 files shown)


vurtun:
Thanks again, interestingly neither clang, gcc or clang analyzer or coverity found the wrong char pointer comparison so great find.

angusholder:
Ah that's strange, I was able to find it using "Clang 3.7 with Microsoft CodeGen" (the version of clang that now ships with VS2015):
1>W:\Game\include\nuklear.h(3116,58): error : comparison between pointer and integer ('const char *' and 'int')
1>          int pattern_repeat = best_letter && pattern_iter != '\0';
1>                                              ~~~~~~~~~~~~ ^  ~~~~

vurtun:
OK then I need to check my makefiles. I use clang 3.7 as well but on linux with -Weverything. Probably have to play around with flags so see if some flags are blocking other. Nonetheless thanks it is always a pleasure to have warning removed or from different platforms and compilers.

PR #66, opened 2016-04-19
Fix some typos.

waywardmonkeys:
This changes a couple of things in code. The only one that might matter is a member ``glyphes`` which is now ``glyphs``.


waywardmonkeys:
I did not correct code which has ``seperator`` rather than ``separator`` as that also modifies a public API I believe.

vurtun:
Thanks! PRs fixing typos are always more than welcome.

PR #67, opened 2016-04-19
Support for OS X

fdb:
This is a proposal for OS X support. 

Currently, I've only implemented this for the GLFW example, but I can update this PR to support the other demos and examples as well.

Here's what it does:
- Update the Makefile to check for OS X and use `-framework OpenGL`.
- Use conditional includes in `main.c`.
- In `nuklear_glfw.c`, define a `NK_SHADER_VERSION` which is set to `#version 400` on OS X and `#version 300 es` otherwise. Version 300 es doesn't work on OS X.

With regards to conditional includes, it seems that GLFW will include the default [OpenGL headers by itself](http://www.glfw.org/docs/latest/build.html#build_include) (this works on OS X at least, haven't tested on Linux / Windows). If this works everywhere we could just remove the conditional include in `main.c`.

### DIFF
--- a/demo/glfw/Makefile
+++ b/demo/glfw/Makefile
@@ -15,7 +15,12 @@ ifeq ($(OS),Windows_NT)
 BIN := $(BIN).exe
 LIBS = -lglfw3 -lopengl32 -lm -lGLU32 -lGLEW32
 else
-LIBS = -lglfw -lGL -lm -lGLU -lGLEW
+	UNAME_S := $(shell uname -s)
+	ifeq ($(UNAME_S),Darwin)
+		LIBS = -lglfw3 -framework OpenGL -lm -lGLEW
+	else
+    LIBS = -lglfw -lGL -lm -lGLU -lGLEW
+	endif
 endif
 
 # Modes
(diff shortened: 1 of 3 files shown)


vurtun:
There is also an old version of OS X, iOS, tvOS which could help: https://github.com/vurtun/nuklear/tree/f1e81a59ea7a4d61e7cb085b273ea3575e28db8f/demo/apple. It uses an old version of this library and I don't have any OS X so I couldn't update it. It may help out with some issues.

fdb:
Thanks!

That approach uses the Cocoa/UIKit layers directly without going through GLFW / SDL / Allegro. That's a valid approach as another backend (we use something similar in-house as well). 

Those shaders have similar conditional code:
```
#if TARGET_OS_IPHONE
    "#version 300 es\n"
#else
    "#version 150\n"
#endif
```

fdb:
I'll see if I can update the rest of the examples / demos tonight.

vurtun:
> That approach uses the Cocoa/UIKit layers directly without going through GLFW / SDL / Allegro
Ah ok I have no experience with apple so thought it could help.

> I'll see if I can update the rest of the examples / demos tonight.
That would be great and more than welcome.

PR #72, opened 2016-04-19
Complete OS X Support

fdb:
This completes OS X support. All demos and backends now compile and run on OS X. I also cursory tested the compile on Ubuntu 15.10, without issues.


vurtun:
Checked it on arch linux as well and everything seems to work. Looks good thank you for your work!

fdb:
You're welcome! Thanks for the great library 👍

PR #73, opened 2016-04-19
correct type definitions without NK_INCLUDE_FIXED_TYPES

ehmry:
HN driveby PR


vurtun:
Thanks great find and fixes.

PR #76, opened 2016-04-19
Spell separator/separated correctly.

waywardmonkeys:


vurtun:
Thanks for the PR. At the moment I don't think it is a good idea to change the API. I will probably bundle it with some other changes and fixes.

PR #81, opened 2016-04-19
Update .gitattributes

1deployer:
porque si


vurtun:
Please at least explain why and don't just push something on to me an expect me to push it!

PR #82, opened 2016-04-20
Use standard C compiler by default

ittner:
Always use the standard compiler as defined by the environment
variable CC. User can set it when a different compiler if required
(e.g. CC=clang make)


vurtun:
I am not 100% sure this is a good idea. My main reasoning for using this Makefile was to switch between compilers to check different outputs in binary or warnings. On the other hand it makes a lot of sense to only use the default compiler in this case.

PR #96, opened 2016-04-20
example/extended: fix blurry font

andreygursky:
Analogously to ca4d5bad8c for the default font ProggyClean.ttf

### DIFF
--- a/example/extended.c
+++ b/example/extended.c
@@ -184,12 +184,10 @@ grid_demo(struct nk_context *ctx)
 
     int i;
     struct nk_panel combo;
-    ctx->style.font.height = 20;
     if (nk_begin(ctx, &layout, "Grid Demo", nk_rect(600, 350, 275, 250),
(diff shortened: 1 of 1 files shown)


vurtun:
There is a reason why I use a font with different heights in that particular demo. So all changes in this PR destroy the meaning of demo `extended.c`. So while the default font looks ugly all extra fonts do not. So what probably would make more sense is to not use the default font at all and instead only use one of the extra fonts and still have text with different height.

andreygursky:
> So while the default font looks ugly all extra fonts do not.

Do you mean if I start like this:
```
./extended ../../extra_font/Cousine-Regular.ttf
```
For me it's still looks blurry:
cousine-regular
Moreover: the letters don't fit vertically and are trimmed (see "g").

Compared to (default 13):
default-13

Other fonts look also blurry for me. Only kenvector_future.ttf and kenvector_future_thin.ttf look good. But kenvector_future.ttf doesn't fit horizontally.

PR #98, opened 2016-04-21
Negative index compilation error tweak

bbrown683:
This was causing some funky compilation errors for MSVC 14, so i just wrapped them into some macro definitions rather than an array since it appears this was mainly for testing and I did not see them being referenced anywhere else.


vurtun:
These typedefs are compiled asserts to make sure that user who do not define `NK_INCLUDE_FIXED_TYPES` are forced to use correct types with correct size. The reasons why you got the compilation error is probably because you either a.) did not define `NK_INCLUDE_FIXED_TYPES` or b.) did not set the correct types with correct size. If it is none of these two and the problem persists feel free to report back and I will look further into it.

bbrown683:
The define appears to have solved all the compilation errors that MSVC reported. Thanks for the information.

PR #102, opened 2016-04-22
Demo with Win32 GDI rendering

mmozeiko:
This adds new `gdi` folder under demo that uses Win32 GDI for rendering. It's based on x11 demo, so it also doesn't implement RECT_MULTI_CLEAR, IMAGE, ARC and ARC_FILLED render commands.

Main loop is structured so it doesn't use 100% of CPU. It waits for Windows message to happen, and only then executes GUI layout and rendering code. If you need continuous update you'll need to use either windows timer to produce periodic messages or remove call to GetMessage and use only PeekMessage functionality.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -663,7 +663,7 @@ NK_API void                     nk_popup_end(struct nk_context*);
 NK_API int                      nk_combo(struct nk_context*, const char **items, int count, int selected, int item_height);
 NK_API int                      nk_combo_seperator(struct nk_context*, const char *items_seperated_by_seperator, int seperator, int selected, int count, int item_height);
 NK_API int                      nk_combo_string(struct nk_context*, const char *items_seperated_by_zeros, int selected, int count, int item_height);
-NK_API int                      nk_combo_callback(struct nk_context*, void(item_getter)(void*, int, const char**), void *userdata, int selected, int count, int item_height);
+NK_API int                      nk_combo_callback(struct nk_context*, void(*item_getter)(void*, int, const char**), void *userdata, int selected, int count, int item_height);
 NK_API void                     nk_combobox(struct nk_context*, const char **items, int count, int *selected, int item_height);
 NK_API void                     nk_combobox_string(struct nk_context*, const char *items_seperated_by_zeros, int *selected, int count, int item_height);
 NK_API void                     nk_combobox_seperator(struct nk_context*, const char *items_seperated_by_seperator, int seperator,int *selected, int count, int item_height);
@@ -19235,7 +19235,7 @@ nk_combo_string(struct nk_context *ctx, const char *items_seperated_by_zeros,
 {return nk_combo_seperator(ctx, items_seperated_by_zeros, '\0', selected, count, item_height);}
 
 NK_API int
-nk_combo_callback(struct nk_context *ctx, void(item_getter)(void*, int, const char**),
+nk_combo_callback(struct nk_context *ctx, void(*item_getter)(void*, int, const char**),
     void *userdata, int selected, int count, int item_height)
 {
     int i;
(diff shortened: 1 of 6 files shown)


dumblob:
Looks good to me. One thing I noticed though is it could be easily overwhelmed in case too many messages are coming from Windows.

Would it be possible to postpone rendering to happen first after some minimum time, whereas during this period the Windows messages would get buffered or just thrown away (depending on their meaning)?

vurtun:
Thanks for your PR. I will test it out and see if everything is working. As it looks to me I probably will have to do something about the style of the code, but that is a small issue.

mmozeiko:
@dumblob Yes, such message limiting capabilities can be implemented. But I don't feel this demo is a good place for that. This demo is about how to implement rendering backend using GDI for nuklear. This is not full-blown application. The code for application should be as simple as possible to make it easier to understand. And all other examples (glfw, SDL, x11) also doesn't limit event processing.

dumblob:
@mmozeiko you're totally right.

After the fast growth of newcomers taking the examples and editing them to reach their own goal of a new application I just got afraid, that nobody will notice this. So to keep things simple, I would advocate for adding a FIXME line saying events should be buffered for some minimum time to the place where the loop waits for new events in all demos. What do you think @vurtun?

nuklear.h
```diff
@@ -663,7 +663,7 @@ NK_API void                     nk_popup_end(struct nk_context*);
 NK_API int                      nk_combo(struct nk_context*, const char **items, int count, int selected, int item_height);
 NK_API int                      nk_combo_seperator(struct nk_context*, const char *items_seperated_by_seperator, int seperator, int selected, int count, int item_height);
 NK_API int                      nk_combo_string(struct nk_context*, const char *items_seperated_by_zeros, int selected, int count, int item_height);
-NK_API int                      nk_combo_callback(struct nk_context*, void(item_getter)(void*, int, const char**), void *userdata, int selected, int count, int item_height);
+NK_API int                      nk_combo_callback(struct nk_context*, void(*item_getter)(void*, int, const char**), void *userdata, int selected, int count, int item_height);
```
paulsapps: Should this be changed in a demo?

mmozeiko: This changed in separate 7536d0d commit.
These things generate warning with VS2015. VS wants function pointers to actually be pointers.

vurtun:
OK I finally got my windows VM and setup working and tested it out. So far it looks good, style fits and everything run smoothly. Good work!

Small nitpick in `nk_gdi_clipbard_paste` and `nk_gdi_clipbard_copy` please try to keep the indentation level as low as possible and return directly on error. I know that Casey does not do this but I personally think it makes things a lot more easy to read.

For future reference please do not bundle PRs with changes to demo code and library. I know you used different commits but there are some problems here for me. First it "hides" (I know that I can still see the commit) changes to the main library under the title "Demo with Win32 GDI rendering" this is manageable issue but still an issue. Furthermore it forces me to reject or accept this commit in total. So for example if I accept the demo but not the changes inside the library you put me in a bad spot and I have to reject everything.
From now on this is a deal breaker for me but this is the first time it happens and the changes to the library are correct so I will let it flow. For any later cases please create two PRs, which would make everything a lot easier for everyone.

@dumblob All demos are meant to be used as a quick jump in point to get something running to test things out and for small apps. But I think you are right it is probably best to at least note limitations   for all demos and fix them, if feasible, in the future .

mmozeiko:
Are you interested in demo with GDI+ rendering? GDI+ can render using antialising, so it will look better than plain GDI.

vurtun:
I never heard of GDI+ before, looks interesting. Your are more than welcome to create a demo for it. In general the more demos the easier it is for someone to just jump in an try out this library.

PR #104, opened 2016-04-22
Added an error message if the display can't be opened

hobbitalastair:
The x11 example currently segfaults if it can't open a display. Detect the failure and print an error message.


vurtun:
Thanks for the PR. I totally missed that `XOpenDisplay` can return NULL.

PR #112, opened 2016-04-24
Demo with Win32 GDI+ rendering 

mmozeiko:
This adds new `gdip` folder under demo that uses Win32 GDI+ for rendering. The advantage over plain GDI is that it gives nice anti-aliased rendering:

screenshot

GDI+ is the rendering used by System.Drawing namespace in .NET Framework. But it can be used from C/C++ without problems. It is included in all Windows versions starting with XP. This GDI+ demo works fine even under wine.

The ugly part is that GDI+ headers are not usable from C, so I copied out all the declarations needed by code and adjusted to work on C.

I changed clipboard functions to have less indentation as you requested in #102. If you want them to look differently let me know and I'll adjust (or feel free to adjust yourself).

### DIFF
--- a/demo/gdip/build.bat
+++ b/demo/gdip/build.bat
@@ -1,0 +1,5 @@
+@echo off
+
+call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
+
+cl /nologo /W3 /WX /O2 /fp:fast /Gm- /Fedemo.exe main.c user32.lib gdiplus.lib /link /incremental:no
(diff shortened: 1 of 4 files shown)


vurtun:
Looks good and runs. Nice piece of work! Especially getting it running under pure C. Thanks for all your porting work so far. Always high quality and fits into what I already provide while also solving the problem on given platform. Even the clipboard functions follow my style now (sorry for the inconvenience).

mmozeiko:
No problem! I can write code in any style, just tell/show me what style it should be :)

PR #113, opened 2016-04-24
Fix issue with adding multiple fonts

SquareWave:
The total range count grabbed from nk_range_count was being used to get the glyph count, which resulted in a buffer over-read. I compiled the SDL demo with MSVC, uncommented a few of the font loading statements, and was getting assertion failures - I'm not positive this is an issue with other compilers but I don't see how it wouldn't be.


vurtun:
Looks good. But it had a warning for `shadowing variable` range_count. I removed one of the to declarations. Other than that great work and thanks for your PR.

PR #114, opened 2016-04-25
Add demo using SDL 1.2. 

glegris:
A lot of platforms are not supported by SDL2 because they are old or nobody has tried to port SDL2 yet. So it can make sense to still use SDL1.2.
Primitives are drawn with the SDL_gfx library.


vurtun:
Looks good, but a few small nitpicks:
In `main.c` it is probably a good idea to remove `nk_color_picker`, since this demo does not support `NK_COMMAND_RECT_MULTI_COLOR`.

I also noticed that there are two different methods used for indentation (tabs as well as spaces). I know there is a lot of debate on the internet, but in this library I use 4 spaces, so it would be nice if this could be fixed.

Another minor thing would be the name. Since it is possible to have both SDL 1.2 + SDL_gfx as well as SDL1.2 + OpenGL and I named the SDL2 + OpenGL version `sdl` it probably makes sense to name it `sdl1.2_gfx` or something related (your call).

The line width used for drawing shape outlines is not really a problem per se since I already allow some platform freedom in other demos. Especially since I don't use line width in this library by default. Other than that there is a comment describing the problem, just like copy & paste so I hope that it is enough (thanks for providing comments in the first place).

Something I am not to happy about is the polygon pointer buffer. In general using `alloca` instead would present itself as a better solution (Even though I am not a big fan) but it would force everyone to use C99 instead of C89. So I am not quite sure. But something that could definitely be done is to move the static declaration into each local function and remove the static. I don't think using a global, persistent storage for this is really necessary or the right solution.

Finally to clarify the `+1` inside `nk_sdl_scissor`. The reason why I used it in the `x11` demo was because it seems to me (I am not 100% sure) that xlib uses a different clipping concept. For example the difference between if a point on the clipping rectangle should be clipped or not. So that is why it is even there. I don't know about the `gdi` demo but I would suspect it is the same issue.

glegris:
I fixed the issues and added support for the `NK_COMMAND_RECT_MULTI_COLOR` command.

PR #125, opened 2016-05-05
Support multiple data sets in a single chart

jviki:
This patch hacks the chart API to support display of multiple data sets in a
single chart instances. Tested briefly with the line chart.

Signed-off-by: Jan Viktorin <iviktorin@fit.vutbr.cz>


vurtun:
Thanks! nice work. I took the liberty and extended and polished everything up a little bit.

PR #126, opened 2016-05-06
s/memset/memcpy/ - typo fix in documentation

hobbitalastair:


hobbitalastair:
Could you please explain why this is wrong?

vurtun:
Oh sorry my bad I though I already fixed that and that's why travis is bitching but you are right I looked at the wrong location. My bad I will pull it.

PR #128, opened 2016-05-08
Close popups by clicking anywhere outside of popup bounds

jimon:
Mostly to recreate combobox behavior from almost every other UI (GTK/MFC/Qt/etc) :+1:


vurtun:
Thanks for the PR, compiles runs and does its job :+1:

PR #129, opened 2016-05-09
Improving window scaling

jimon:
In most UI frameworks scaling "attaches" mouse cursor to scale pivot, so if even mouse cursor leaves the pivot (in case if scaling restrictions don't allow pivot to follow cursor) - scaling is not canceled, and pivot will follow again when cursor position will generate valid window size.

Previous nuklear behavior was to cancel scaling at moment when cursor moves outside of scaling pivot, which is not natural for many users. Plus also previous implementation used incremental delta integration which generates inaccurate inputs if the mouse leaves and then re-enters the app window.

I've changed the implementation to have scaling state with initial size so we can apply total accumulated delta to always have correct good looking results.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -2436,6 +2436,11 @@ struct nk_property_state {
     int state;
 };
 
+struct nk_scaling_state {
+    struct nk_vec2 start_size;
+    int active;
+};
+
 struct nk_window {
     unsigned int seq;
     nk_hash name;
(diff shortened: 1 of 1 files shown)


vurtun:
Thanks for another great PR. I am not quite sure this is the correct way of achieving the desired result but I think it is good enough. But this PR introduced a few ANSI C errors, warnings and style issues that should be fixed now.

jimon:
@vurtun what would be the correct way then? Because I'm planning to do similar thing to the scrolling :)

vurtun:
It could be that I don't understand what you are doing, but scrollbar already has mouse grabbing to allow scrolling even if the mouse is outside the scrollbar, just like slider.

PR #130, opened 2016-05-15
Some checks

romku96:
Added some checks


romku96:
There is exclusive zero check in `nk_inv_sqrt`, but 0 is acceptable too

PR #152, opened 2016-05-24
Allow static windows position update.

Tirus42:
Currently any position change is ignored.
Its a problem if you want to create a gui and rescale the draw area during runtine.
This patch will allow windows with not NK_WINDOW_MOVEABLE set to update the position.


vurtun:
Sorry but while correct I don't think it is a good idea. I think adding multiple meanings to one argument with different side effects depending on a single flag is not good API design and is confusing. Especially since I already provide functions to access and modify window position and size by `nk_window_set_bounds`, `nk_window_set_position` and `nk_window_set_size`.

PR #154, opened 2016-05-27
fixed 'unsigned short' overflow when drawing widget text with font size > widget height

ytsarko:
When `f->height` is greater than `b.h` then resulting `label.h` will become negative and inside call of `nk_draw_text` in expression `cmd->h = (unsigned short)r.h;` resulting h will become **USHRT_MAX**.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -9838,7 +9838,9 @@ nk_font_atlas_end(struct nk_font_atlas *atlas, nk_handle texture,
     }
     for (i = 0; i < atlas->font_num; ++i) {
         atlas->fonts[i]->texture = texture;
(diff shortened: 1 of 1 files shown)


ytsarko:
Sorry, the second commit (de91bd6) got here accidentally. It really should be a separate PR.

vurtun:
Thanks for your PR. While fixing the problem it also let to me fixing another bug I found because of it inside the `xlib` demo. :+1: 
As for the second commit I let it slide since the changes are not that significant (as in complex for me to follow) otherwise please create two PRs in the future.

ytsarko:
> As for the second commit I let it slide since the changes are not that significant (as in complex for me to follow) otherwise please create two PRs in the future.

Sure, I`ve mistakenly added that second commit to a PR.

PR #164, opened 2016-06-13
Emscripten support with glfw3 (and shaders) demo at www.cwkx.com/nuklear

cwkx:
Hi, I added Emscripten support with glfw3 (and shaders), with a demo at www.cwkx.com/nuklear

- to compille, use "emcc main.c -o index.html -s USE_GLFW=3" or use "node build.js" once emscripten is installed and added to path.
- You can get index.js down to 450kb with -Oz but this requires a webserver for the index.html.mem whereas with -O0 it runs out of the box
- the main changes to make it work were some rewrite of the shaders, and replacement of glMapBuffer with glBufferSubData, and some restructuring (adding meta to device struct) as VAOs aren't supported by WebGL.


lieff:
Additional compiler switch -I../.. is needed for #include "nuklear.h".

lieff:
Works great!

dumblob:
Thanks, it's nice to see Nuklear working in a web browser!

By the way, do you also experience window grab exchange or even both-windows-at-the-same-time-grab when quickly moving one window back and forth while ensuring the mouse pointer is crossing another window's title bar (disregarding whether the grabbed window is overlapping or underlapping)?

@vurtun do you happen to have any idea why this could happen?

paulsapps:
Mouse wheel scrolling seems to be inverted?

dumblob:
@cwkx could you please update the demo on www.cwkx.com/nuklear (as the bug mentioned in https://github.com/vurtun/nuklear/pull/164#issuecomment-227692547 should be fixed by https://github.com/vurtun/nuklear/commit/2b1c2ca5406714e3e673fa5fcb28224ef517d403 )?

dumblob:
@cwkx I can see that you're working hard on the 3D curve paper, but could you please state whether you want to further develop/maintain the emscripten demo? @vurtun decided to remove it in https://github.com/vurtun/nuklear/commit/cae6005015383cf70f8f6db653b440e1e6ada613 because of missing response.

PR #167, opened 2016-06-17
Moving some very useful macros to library interface part

ytsarko:
Some very useful macros (especially NK_MIN/MAX, NK_CLAMP, NK_BETWEEN, NK_INBOX, NK_ALIGNOF and others) are moved from NK_IMPLEMENTATION part of the header to interface part.

Having these in interface will facilitate in creating of nuklear backends when one needs to perform calculations of rectangle intersections, allocating memory in nk_buffer (**nk_buffer_push**, for example, accepts alignment of an object) range checks and others.


ytsarko:
Hi!
Are there any chances it will be merged?

vurtun:
Hey sorry this took a little longer I don't have much time at the moment.
I don't think there is any good reason to make these public. As far as I would say everyone
who wants to use them can just move them into the header part while those that don't want to clutter their namespace by default. So I am not 100% sure and tend to keep it private or are there any other advantages of having them public.

ytsarko:
Hi!
> I don't think there is any good reason to make these public. As far as I would say everyone
who wants to use them can just move them into the header part...

Actually, this PR is aimed to reduce those number of cases. This will allow minimize the number of changes between upstream nuklear.h and one used in the project and as a side effect simplify syncing with upstream and other maintenance.

> ... don't want to clutter their namespace by default

As far as I can tell, all the moved stuff begins with NK_ prefix, so I think it hardly can hit anyone. If someone has chosen to use such a prefix for his own code then he will anyway have troubles using nuklear because of other functions/types/macros that are already in the interface part of the library.

PR #172, opened 2016-06-24
horizontal scrollbar fixes

inolen:
While working on a texture preview window containing a horizontally scrolling window of images, I hit a few edge cases with horizontal scrollbars.

First, my window didn't container a scaler so they panel's height was calculated incorrectly. The old calculation used scaler_size (16) when either scrollbars or scalers were enabled, as opposed to the max of either scaler_size (16) or scrollbar_size.y (10). Since my window was aligned to the bottom of the screen there was a 6 pixel gap between container and the bottom of the screen due to this.

Second, the horizontal scrollbars used the scrollbar_size.x to determine their final position, as opposed to scollbar_size.y. Being that my window didn't have any vertical scrolling I had set scrollbar_size.x to 0, which then causes the horizontal scrollbar to be position incorrectly.

I've attached a before and after of the use case.

Before (scrollbar is completely not visible):
screenshot from 2016-06-23 22-44-34

After (bright green is the scrollbar background):
screenshot from 2016-06-23 22-43-09

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -15633,10 +15633,16 @@ nk_panel_begin(struct nk_context *ctx, const char *title)
     }
 
     /* calculate window footer height */
-    if (!(win->flags & NK_WINDOW_NONBLOCK) &&
-        (!(win->flags & NK_WINDOW_NO_SCROLLBAR) || (win->flags & NK_WINDOW_SCALABLE)))
-        layout->footer_h = scaler_size.y + style->window.footer_padding.y;
-    else layout->footer_h = 0;
+    layout->footer_h = 0;
+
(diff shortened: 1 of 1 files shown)


vurtun:
Thanks for the great PR! Good explanation what is wrong (even with pictures!) and what should be :+1: .Small nitpick please use the correct code style. Single `if` statements do not have `{}` and I use 4-Spaces for Tabs. I fixed it myself but it would be great if I don't have to do this myself for every PR. Other than that A+ commit.

inolen:
@vurtun Sorry I missed the style differences, thanks for fixing things up!

It may be worth investigating clang-format and seeing if there's a happy medium you can achieve with its options that people can run before commit.

PR #181, opened 2016-07-12
Fix compilation when NK_INCLUDE_COMMAND_USERDATA is defined

Spasi:


vurtun:
Thanks for the PR. While it is correct that this is an error your fix is overly complex. Instead of  
```c
(&ctx->draw_list)->userdata = cmd->userdata;
``` 
you  just could have fixed it with:
```c
ctx->draw_list.userdata = cmd->userdata;`
```
I went ahead and fixed it with the second version.

PR #184, opened 2016-07-15
Enhancement: 'nk_buffer_push' function now returns address

ytsarko:
Previously, **nk_buffer_push** function returned void. With this commit it returns address of item
which was added to buffer. One can use this address to get a pointer to that item inside the buffer (for example to modify its fields or pass this pointer to another code).

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -10000,7 +10000,6 @@ nk_input_key(struct nk_context *ctx, enum nk_keys key, int down)
     NK_ASSERT(ctx);
     if (!ctx) return;
     in = &ctx->input;
-    if (in->keyboard.keys[key].down == down) return;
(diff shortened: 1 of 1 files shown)


vurtun:
Thanks for the PR. But if you _really_ want the pointer please use `nk_buffer_alloc` instead. The reason why I did not return the new memory from `nk_buffer_push` is because the pointer access behavior is undefined after the next call to `nk_buffer_push` since the buffer could reallocate. `nk_buffer_alloc` has the same behavior that is why it is private. If you know what you are doing you can make it public and use it.

ytsarko:
You`re right. I absolutely forgot about buffer reallocs which render pointer at buffer contents useless.

PR #192, opened 2016-08-03
Fixed multiple crashes in font handling

rokups:
Crashes happen when more than one font is present. First issue is only first font being initialized over and over and second one is `nk_zero` clearing `font->next` value which also prevents initialization of any fonts but first one.

As a bonus i snuck in compile error fix. c++ templates can not have external linkage. Whatever compiler is accepting that should be fixed ;)

P.S. i couldnt get font-awesome to work anyhow, displays question mark if i try to use

        nk_style_set_font(nk, &fa->handle);
        nk_button_label(nk, "\xEF\x80\x80");
        nk_button_label(nk, "\uF04B");

Any tips to get that working would be appreciated ;)

P.P.S. sorry for irrelevant space removals from line ends, blame IDE. If you wish i can clean commits up and get rid of them.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -8917,7 +8917,7 @@ nk_font_bake_pack(struct nk_font_baker *baker,
     /* setup font baker from temporary memory */
     for (config_iter = config_list; config_iter; config_iter = config_iter->next) {
         const struct nk_font_config *cfg = config_iter;
-        if (!nk_tt_InitFont(&baker->build[i].info, (const unsigned char*)cfg->ttf_blob, 0))
+        if (!nk_tt_InitFont(&baker->build[i++].info, (const unsigned char*)cfg->ttf_blob, 0))
             return nk_false;
     }
 
@@ -9263,7 +9263,6 @@ nk_font_init(struct nk_font *font, float pixel_height,
         return;
 
     baked = *baked_font;
-    nk_zero(font, sizeof(*font));
     font->info = baked;
     font->scale = (float)pixel_height / (float)font->info.height;
     font->glyphs = &glyphs[baked_font->glyph_offset];
--- a/nuklear.h
+++ b/nuklear.h
@@ -1224,7 +1224,7 @@ NK_API void nk_textedit_redo(struct nk_text_edit*);
     The final approach if you do not have a font handling functionality or don't
     want to use it in this library is by using the optional font baker. This API
     is divided into a high- and low-level API with different priorities between
-    ease of use and control. Both API's can be used to create a font and 
+    ease of use and control. Both API's can be used to create a font and
     font atlas texture and can even be used with or without the vertex buffer
     output. So it still uses the `nk_user_font` struct and the two different
     approaches previously stated still work.
@@ -2533,6 +2533,8 @@ struct nk_context {
     (type*)((void*)((char*)(1 ? (ptr): &((type*)0)->member) - NK_OFFSETOF(type, member)))
 
 #ifdef __cplusplus
+}
+
 template<typename T> struct nk_alignof;
 template<typename T, int size_diff> struct nk_helper{enum {value = size_diff};};
 template<typename T> struct nk_helper<T,0>{enum {value = nk_alignof<T>::value};};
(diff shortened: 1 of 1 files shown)


vurtun:
Big thank you for your PR and your hard work. I will pull and test it as soon as I get home and I will try out font-awesome as well and see if I can get it work. Interestingly I had to fix the template error today as well but you beat me to it.

vurtun:
I pushed the changes but was not able to get a clean merge so I just wrote a small commit. I will close this pull request. Sorry that I couldn't directly do the merge but I fcked up handling git :frowning:

rokups:
Its ok, as long as fixes are in ;)

Hey is there any place i could possibly bother people with questions about this lib? If not, maybe consider a gitter chatroom or something?

vurtun:
In general I would say just feel free to create an issue about anything and I will try to answer or if possible write a small demo showcasing how to do things. Most of the time these questions help out others as well and are a net positive all around.

vurtun:
Small note this is not really to appropriate place to write this but you asked for font-awesome and I think I got it to work:

```c
    struct nk_font_config config = nk_font_config(13);
    const nk_rune font_awesome_ranges[] = {
        0x0020, 0x00FF,
        0xF000, 0xF299,
        0
    };
    config.merge_mode = nk_false;
    config.pixel_snap = 0;
    config.oversample_v = 1;
    config.oversample_h = 1;
    config.size = 14;
    config.coord_type = NK_COORD_UV;
    config.spacing = nk_vec2(0,0);
    config.fallback_glyph = '?';
    config.range = font_awesome_ranges;
    struct nk_font *awesome = nk_font_atlas_add_from_file(atlas, "../../../extra_font/fontawesome-webfont.ttf", 13, &config);
```
Like you can see it requires you to create a `struct nk_font_config` struct and set the appropriate font ranges. I am not 100% how this font works but I get symbols for ASCII characters as well so would be cool if you could test it and check if it is correct.

rokups:
This font provides common icons, thats why its font-awesome ;)) I can see it work somewhat now, but it is not working as expected. For example:

    nk_style_set_font(nk, &fa->handle);
    nk_button_label(nk, "\uf003");
    //nk_style_set_font(nk, &font_default->handle);

This is supposed to display button with [envelope](http://fontawesome.io/icon/envelope-o/) icon (font-awesome says it's utf-8 codepoint is f003). However i get a hammer:
hammer
Readme says nuklear supports utf-8, but maybe i am using it in a wrong way?

Another problem with code snippet above is if i uncomment last line it makes button display question mark instead. Maybe i am using fonts wrong here? I tried button without any `nk_style_set_font` calls hoping that nuklear will pick the right font based on range glyph is in - that did not work either and i still see a question mark.

dumblob:
>This is supposed to display button with envelope icon (font-awesome says it's utf-8 codepoint is f003). However i get a hammer...

Just an immediate thought - are you aware of the [utf8 NFC and NFD](https://en.wikipedia.org/wiki/Unicode_equivalence#Normal_forms ) differences? I did not check how it is handled in Nuklear though.

rokups:
No i am not. I heard terms few times in the past and assumed its some low level stuff programmers usually dont bump into.

vurtun:
@rokups First of 
> Another problem with code snippet above is if i uncomment last line it makes button display question mark instead. Maybe i am using fonts wrong here?

This was a bug and I just fixed and pushed a commit for it.

> Readme says nuklear supports utf-8, but maybe i am using it in a wrong way?

I tested it step for step and the unicode decoding works correctly and I get the correct codepoint from `"\uf003"` which is `61443`. I also checked the glyph querying function and that as well is correct. Which only leaves the glyph renderer or rather the font format. I don't know exactly at the moment what font-awesome does but it is not standard behavior (maybe what @dumblob said, different format).

rokups:
I think i forgot to test font swapping fix you did. So now i did and guess what - it all works:
mail-button
However as you can see there is problem with bottom of icon being cut-off. Changing font size or button size has no effect. Maybe it is related to issue #204?

Anyhow thanks for the help with this, now nuklear can use vast amount of fancy icons with zero effort ^_^

PR #197, opened 2016-08-04
Bugfixes

rokups:
Two bugfixes:

* Zero-terminate strings modified by `nk_edit_(buffer|string)`
* External linkage for implementation. Without it having implementation part enabled in `.cpp` file does not work.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -2584,6 +2584,9 @@ template<typename T> struct nk_alignof{struct Big {T x; char c;}; enum {
  * ===============================================================
  */
 #ifdef NK_IMPLEMENTATION
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 #ifndef NK_POOL_DEFAULT_CAPACITY
 #define NK_POOL_DEFAULT_CAPACITY 16
@@ -20276,4 +20279,7 @@ NK_API void
 nk_menu_end(struct nk_context *ctx)
 {nk_contextual_end(ctx);}
 
+#ifdef __cplusplus
+}
+#endif
 #endif


vurtun:
Thanks for the PR, but there are some problems.
1.) Please divide PRs up so it is easier for me to accept or deny single changes instead of bundles of changes.
2.) There is a reason why I did not zero terminate string with `nk_edit_buffer` and `nk_edit_string` since I do not want to enforce a certain string format on other people. However you are right and I will create two additional API calls that will do the zero terminating.
3.) The part about exernal linkage I do not get. I directly include it in a .cpp project just fine without any problems under Visual Studio 2015. Does it not work for you and was I just lucky? Under normal circumstance the `extern C` is only to prevent C++ name mangling so I am not 100% sure this is the right thing to do.

rokups:
Okay sorry, did not want to spam you too much. Ill get rid of zero-terminating mod. Lack of documentation really bites us in the ass here.. See how i thought it was a bug.

I am surprised that including `nuklear.h` in `.cpp` for implementation works, because function declarations are wrapped in `extern "C"` and this they get c linkage without name mangling. However implementations are outside of `extern "C"` and if they are included in `.cpp` file they should get c++ linkage with name mangling. That results in unresolved externals. If this works on msvc then maybe compiler is doing something non-standard here..

vurtun:
> That results in unresolved externals. If this works on msvc then maybe compiler is doing something non-standard here.

I tested it on msvc, gcc and clang and it never caused any problems, I also looked at other single header libraries and none else has it. Furthermore the standard specifically states:

> A function can be declared without a linkage specification after an explicit linkage specification has been seen; the linkage explicitly specified in the earlier declaration is not affected by such a function declaration.

So while you are not wrong in that it is possible to do, I am not required to do it and I would prefer the leave out everything not really required.

rokups:
That makes sense. Ill try to dig around see what else could have caused this. Must be gremlins in my code i guess. Thanks for taking time to look into this.

PR #198, opened 2016-08-04
Added window flag NK_WINDOW_SCALE_LEFT

rokups:
Added window flag `NK_WINDOW_SCALE_LEFT` which puts scaling handle to the left-bottom corner of the window and window scales to the left side instead of right one.

Motivation for this change is usecase when windows are used as panels attached to the right side of screen. Naturally expanding them is not possible because mouse goes out of screen after 10 pixels or so.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -575,7 +575,8 @@ enum nk_panel_flags {
     NK_WINDOW_NO_SCROLLBAR      = NK_FLAG(5), /* Removes the scrollbar from the window */
     NK_WINDOW_TITLE             = NK_FLAG(6), /* Forces a header at the top at the window showing the title */
     NK_WINDOW_SCROLL_AUTO_HIDE  = NK_FLAG(7), /* Automatically hides the window scrollbar if no user interaction */
-    NK_WINDOW_BACKGROUND        = NK_FLAG(8) /* Always keep window in the background */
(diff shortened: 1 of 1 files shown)


nuklear.h
```diff
@@ -16865,17 +16866,30 @@ nk_panel_end(struct nk_context *ctx)
         scaler.w = scrollbar_size.x;
         scaler.h = scrollbar_size.y;
         scaler.y = layout->bounds.y + layout->bounds.h;
-        scaler.x = layout->bounds.x + layout->bounds.w + panel_padding.x;
+
+        if (layout->flags & NK_WINDOW_SCALE_LEFT) {
+            scaler.x = layout->bounds.x - panel_padding.x / 2;
```
rokups: @vurtun rebased this PR on to master but need you to verify this line. I have no idea if i did it correct here. Basically this is a guess that worked.

rokups:
I understand my PR might be not acceptable. Especially since i was not sure 100% on correctness of changes. If this is the case why it went ignored by you feel free to close it. Lack of attention is also why i stopped maintaining it. But it would be still cool to have `NK_WINDOW_SCALE_LEFT` as it would allow us to use windows as sidebars on the right side of screen. Very useful for things like game editors.

vurtun:
Thanks for your PR, I indirectly merged it in commit: 34d0fcd43dff6eceddf2dbe7604e2e4a428a97f9 since it had some merge conflicts.

PR #200, opened 2016-08-04
Property widgets use native numbers instead of converting them to float.

rokups:
And added property widgets for doubles.

Reason for this change is me bumping into precision loss issue when making GUI editor for game engine. Integers loose some precision when converting to floats. For example `M_MAX_INT` value gets changed and delta is in couple-hundreds. I did it mainly for ints but added doubles because they sometimes are needed too so why not..

This is pretty big patch and aside from widget functions other most invasive ones are:

Converted following functions to use doubles instead of floats: `nk_strtof` (renamed to `nk_strtod` and made it use doubles instead. Ofc added `nk_strtof` which does some typecasting and uses `nk_strtod`), `nk_ifloor`, `nk_iceil`, `nk_log10` and `nk_pow`.

Wrote `nk_strtoi`, `nk_strrev_ascii`, `nk_itos` functions which should be carefully verified by independent eyeballs.

I realize it is bit invasive change that may or may not be accepted. Using doubles instead of floats means a performance hit. No idea if it will be noticeable ( i think most likely not)? Is possible performance hit worth accuracy when dealing with numbers? Did i mess something up? Therefore im creating this PR so we can discuss. I personally think that we really could benefit from widgets that do not artificially mess up values.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -726,8 +726,10 @@ NK_API int                      nk_color_pick(struct nk_context*, struct nk_colo
 /* Widgets: Property */
 NK_API void                     nk_property_float(struct nk_context *layout, const char *name, float min, float *val, float max, float step, float inc_per_pixel);
 NK_API void                     nk_property_int(struct nk_context *layout, const char *name, int min, int *val, int max, int step, int inc_per_pixel);
+NK_API void                     nk_property_double(struct nk_context *ctx, const char *name, double min, double *val, double max, double step, double inc_per_pixel);
 NK_API float                    nk_propertyf(struct nk_context *layout, const char *name, float min, float val, float max, float step, float inc_per_pixel);
 NK_API int                      nk_propertyi(struct nk_context *layout, const char *name, int min, int val, int max, int step, int inc_per_pixel);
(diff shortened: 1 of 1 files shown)


vurtun:
Thanks for the PR. I took a look and I probably have to take and rewrite it a little bit to make it fit. But before that I want to ask the reason for double. You said precision problems but are these caused by the conversion from string to float (which has quite bad conversion precision) or actually by float limitations. If it is because of string conversion double will not really help, rather I would have to provide a way to include the highly accurate standard library `strtod` function (which is something I probably should do either way). If it is actually about float limitations (which would surprise me) then I will add the double version based on your implementation.

rokups:
Doubles suffer from converting to floats **a lot**, and not only part after floating point. Main part of number becomes inaccurate with not that big float numbers while it remains accurate with really large doubles. I suppose it cant hurt to have a switch that enables api version with doubles at cost of using `strtod`. As long as people can opt out it would be net-gain.

vurtun:
I will close this PR since it will get conflicts while I am rewriting my own version for now but I promise I will implement a solution solving the problem.

vurtun:
So importing this PR should be complete now. I took must of your implementation but changed some style and naming issues. I am not 100% sure using double is the best idea since I wanted to make sure to only use float to support more platforms but I will stick with double until somebody brings it up.

PR #216, opened 2016-08-23
add missing CLFW_PRESS check for GLFW_KEY_RIGHT_CONTROL

kybin:
It seems CLFW_PRESS check for GLFW_KEY_RIGHT_CONTROL is missing in examples. 
Added them.


vurtun:
Thanks for the PR. Seems like a missed that.

Issue #215, opened 2016-08-21
Text Input for edit_string/property/et.al, virtual keyboards, and SDL


ewmailing:
I'm banging my head on how to handle with text input via SDL & Nuklear in a way that works across all platforms. The big problem is mobile (e.g. iOS, Android) where there is no physical keyboard and a virtual on-screen keyboard must be used. However there are subtle related issues even for desktop. Even though I'm talking about SDL, I suspect this is a common problem across all libraries.

To jump to the end, I'm thinking Nuklear needs a callback for when a edit field first enters edit mode so the correct SDL (or whatever) functions can be invoked to activate the appropriate input mode.

So more specifically, SDL offers these 3 really important functions for text input.
SDL_StartTextInput()
SDL_StopTextInput()
SDL_IsTextInputActive()

For all platforms, this will turn or off SDL's Text Input (and Text Editing) modes. This was intended to handle more complicated text scenarios where multiple keypresses may make a glyph. The current SDL adaptor for Nuklear depends on this being activated. As far as I can tell, this mode is by-default-on for most desktops, but default-off for mobile. But it is worth noting that users may toggle this for their own needs, presumably something where they only care about raw key presses, maybe suitable for standard game controls via keyboard. So this is one reason why I think Nuklear needs a callback, so SDL_StartTextInput() can be called at the right time if the rest of the app needs it off.

For platforms with virtual keyboards (i.e. mobile), SDL_StartTextInput() and SDL_StopTextInput() automatically bring up and hide the on-screen keyboard. So being able to call this at the right time is essential for getting keyboard input to work at all.

So basically, the workflow needs to be:
1. User mouse clicks or touches the editable field
2. Some callback gets invoked so SDL_StartTextInput() can be called.
3. User types in stuff and is sent properly to Nuklear (the (evt->type == SDL_TEXTINPUT) case in the current adaptor).
4. User leaves the textfield editing mode somehow (clicking/touching out).
5. Another callback gets invoked so SDL_StopTextInput() can be called as appropriate. (Depending on needs, this could be left on.)

Notes:
A. It is worth noting that we must only call SDL_StartTextInput() at the beginning of the edit, and not every frame. The reason is that for virtual keyboards, there is a dismiss button which is not a regular button that sends events. It is solely responsible for dismissing the keyboard. So if we accidentally keep calling SDL_StartTextInput() every frame, every time the user tries to close the keyboard so they can try to get out the text editing mode, the keyboard will be forced back up and the user will be trapped.

[...]

PR #222, opened 2016-08-26
Do not insert a backward delete

KalenAnson:
Not sure if this is something that is happening on several platforms, but my Mac keyboard sends unicode '127' as 'backspace' or the backward delete key.

Previously this event was being handled both as a valid text entry and as a text edit event. So the character was being inserted into the text edit string and then deleted, rendering the delete event non-functional.

Let me know if this needs to be handled elsewhere.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -10861,6 +10861,10 @@ nk_textedit_text(struct nk_text_edit *state, const char *text, int total_len)
     if (!glyph_len) return;
     while ((text_len < total_len) && glyph_len)
     {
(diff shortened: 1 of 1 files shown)


KalenAnson:
Not sure if this is related to #215 but I have been seeing this problem for a while on an app I have been working on.

vurtun:
Thanks for the PR. Mac seems to be the biggest trouble maker so far. Hopefully this change does not have any side effects I cannot for see but I can't think of any circumstance it would right now.

PR #244, opened 2016-09-18
fix compiler warning in switch statements

flederfalter:
Clang emits by default a compiler warning if not all enumeration values are handled in a switch. I also added an assert just in case.


vurtun:
Thanks for the PR fixed warnings in b0e2dba52cad9f3a796bcc8db501be58b194b95b.

PR #245, opened 2016-09-18
fixed C89 compatibility brakeage

flederfalter:
The function `vsnprintf` is only available since C99 (see: http://en.cppreference.com/w/c/io/vfprintf, also compiler emits warning with -std=c89). However there is no way to check the buffer size with C89 standard library to prevent possible overflow. I tried to hack it first, but no use.
There are only two solutions: implementing own function or rely safety on an assert. I decided the easier option for now.

Also` vsprintf` is dependent on ` <stdio.h>` so I added `NK_INCLUDE_STANDARD_IO` in all the define conditions.


vurtun:
Thanks for the PR. While I agree that this is a problem and even a bug I think your solution is not enough since I don't want to use an unsafe function by default. I fixed the issue in a little bit different way in commit: b0e2dba52cad9f3a796bcc8db501be58b194b95b. My solutions could  screw me at a later point of time but for now I think it is sufficent.

PR #259, opened 2016-10-17
Correct NK_MEMCOPY optional define to NK_MEMCPY

ettoretorti:
The documentation lists NK_MEMCOPY as the optional macro to be defined to provide a custom 'memcpy', but the rest of the file uses NK_MEMCPY.


vurtun:
Thanks for the PR. Seems like this typo slipped through the cracks.

PR #260, opened 2016-10-22
Improved and fixed example/Makefile.

dankan1890:


vurtun:
Thanks for the PR. I tested it and it worked like a charm.

PR #261, opened 2016-10-23
Add x11_cairo demo renderer.

ventosus:
Add cairo demo renderer for X11.

Via the switch <NK_CAIRO_GL>, cairo will either render to an OpenGL surface
or via xlib/xrender directly.

### DIFF
--- a/demo/x11_cairo/Makefile
+++ b/demo/x11_cairo/Makefile
@@ -1,0 +1,15 @@
+# Install
+BIN = demo
+
+# Flags
+CFLAGS = -std=c89 -pedantic -O2 -I/usr/include/cairo
+
+SRC = main.c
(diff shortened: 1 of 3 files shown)


vurtun:
Thanks for the PR but please update your version to the new nuklear version and fix all code style issues and for demos try to fit to all the other demos (for example your `main.c` is missing optional style) before creating a PR. I know that everyone believes that his/her style is the one and only but I need to preserve *one* style for the whole project. I also tried to run it and it froze on me, not sure if I did something wrong while updating so I would probably test it again if updated.

dumblob:
@ventosus do you have somewhere an updated version of the Cairo backend?

ventosus:
@dumblob, not yet, but it's on the TODO list :)

PR #270, opened 2016-11-01
bugfix: nk_combo_separator showed incorrect item length and broken it…

nsawa:
I think, there is a bug in nk_combo_separator.
I tried to fix it.
See below for details.
Thanks.

test_program
before_fixing
after_fixing


vurtun:
Look good thanks for the PR. Everything works correctly now, great work.

PR #271, opened 2016-11-01
remove useless variable which causes wild  pointer error

sillyousu:


vurtun:
Thanks for the PR, seems like I missed a panel to remove.

PR #273, opened 2016-11-03
fix: nk_value_color_byte() shows incorrect text

nsawa:
I think, there is a bug in nk_value_color_byte.
I tried to fix it.
See below for details.
Thanks.
fix__nk_value_color_byte__test_program
fix__nk_value_color_byte__before_fixing
fix__nk_value_color_byte__after_fixing


vurtun:
Thanks for your work, look good.

PR #274, opened 2016-11-03
fix: nk_hsva_f() lacked processing for quadrant 3

nsawa:
I think, there is a bug in nk_hsva_f.
I tried to fix it.
See below for details.
Thanks.
fix__nk_hsva_f__before_fixging
fix__nk_hsva_f__after_fixging


vurtun:
Thanks for the PR I tested and everything works correctly now. Thanks for your work.

PR #276, opened 2016-11-05
fix: nk_value_color_hex() showed incorrect value.

nsawa:
I think, there is a bug in nk_color_hex_rgba and nk_color_hex_rgb.
I tried to fix it.
See below for details.
Thanks.
test_program
before_fixing
after_fixing


vurtun:
Thanks again for finding and fixing the problem: A+.

PR #277, opened 2016-11-10
Fix edit cursor movement and text overflow

keharriso:
These changes make cursor movement more natural for edit widgets. Up and down behave like you'd expect them to even at the end of a line or with empty lines. It also sets the scissor properly for inactive edit widgets, preventing text overflow.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -11637,31 +11637,28 @@ nk_textedit_find_charpos(struct nk_text_find *find, struct nk_text_edit *state,
     int z = state->string.len;
     int i=0, first;
 
+    nk_zero_struct(r);
     if (n == z) {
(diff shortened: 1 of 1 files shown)


vurtun:
Hey and thanks for the commit. Honestly I tried your PR and I am not quite able to make out what
you changed. For me if I try the my version or yours I get the same result. Got you explain your changes in more detail. In addition you used `={0}` which does not exist in C++ and has to be exchanged with a `nk_zero` call.

keharriso:
The PR solves two issues:
* Text no longer overflows beyond the boundaries of the edit field.
* Moving the cursor up/down when the current x position is beyond the length of the next line causes the cursor to either move down two rows instead of one (when pressing down) or to move to the beginning of the current line (when pressing up).

issue277

PR #280, opened 2016-11-11
fix: using nk_layout_row_begin() with NK_STATIC makes incorrect layout

nsawa:
Using nk_layout_row_begin() with NK_STATIC makes an incorrect layout.
I tried to fix it.

At least, this fixing makes correct result in this example.
But I'm not sure whether this fixing is always correct...

See below for details.
Thanks.
1_test_program
2_before_fixing
4_after_fixing


vurtun:
Thanks for another great PR.

PR #284, opened 2016-11-15
fix: key reaction timing in Windows demos

nsawa:
There is a problem in the gdi, gdip and d3d11 demos.
The key reaction is occured at the timing of the control keys (cursor key, enter key, home key, etc) are released.
The cause is that the 'down' flag is reversed.

In nuklear_gdi.h, nuklear_gdip.h and nuklear_d3d11.h:
>int down = (lparam >> 31) & 1;

I think the following is correct:
>int down = (lparam >> 31) & 1 ? 0 : 1;

Because [MSDN](https://msdn.microsoft.com/ja-jp/library/windows/desktop/ms646280(v=vs.85).aspx) says:
>lParam Bits 31
>The value is always 0 for a WM_KEYDOWN message.
>The value is always 1 for a WM_KEYUP message.
>The value is always 0 for a WM_SYSKEYDOWN message.
>The value is always 1 for a WM_SYSKEYUP message.

See below for details.
Thanks.

test_program
before_fixing
after_fixing

### DIFF
--- a/demo/d3d11/nuklear_d3d11.h
+++ b/demo/d3d11/nuklear_d3d11.h
@@ -205,7 +205,7 @@ nk_d3d11_handle_event(HWND wnd, UINT msg, WPARAM wparam, LPARAM lparam)
     case WM_SYSKEYDOWN:
     case WM_SYSKEYUP:
     {
-        int down = (lparam >> 31) & 1;
+        int down = (lparam >> 31) & 1 ? 0 : 1;
         int ctrl = GetKeyState(VK_CONTROL) & (1 << 15);
 
         switch (wparam)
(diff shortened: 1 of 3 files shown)


dumblob:
@nsawa thanks for the catch. I would though prefer `int down = ! ((lparam >> 31) & 1);` over `int down = (lparam >> 31) & 1 ? 0 : 1;` because of readability and actually even performance (ternary operator will highly probably be translated to a jump, which is slow compared to negation). What do you think?

nsawa:
Thank you for the comment. I think so, too. I changed the code.

paulsapps:
No one probably cares but both produce same machine code when optimizations are enabled.

dumblob:
@paulsapps I definitely wouldn't dare to state, that the ternary operator variant is always compiled into the same code as the negation variant on **all** compilers (including unknown embedded ones) supporting ANSI C and a CPU/SoC architecture capable of running Nuklear when compiler optimizations are enabled. Therefore I do care :wink:.

paulsapps:
I'm sure there are plenty of old buggy/home made compilers that would produce a dogs dinner of either ways of writing it :), my point was just that anything that isn't antique handles it as you'd expect.

PR #286, opened 2016-11-16
fix: popup border is not shown when using the default style settings

nsawa:
Popup border is not shown in the default style settings.
I think, the initialization of popup border style is missing.
Specifically, although nk_popup_begin() specifies the NK_WINDOW_BORDER flag, nk_style_from_table() hasn't initialized win->popup_border_color and win->popup_border.

See below for details.
Thanks.

before_fixing

after_fixing


vurtun:
Thanks for the PR good work.

PR #306, opened 2016-12-26
fixed relative path of demo fonts in d3d11 sample

juliuszint:
The sample code that loads different fonts into the context traversed back once to much and the Roboto font name missed a o at the end.


vurtun:
Thanks for the PR, looks good!

Issue #305, opened 2016-12-26
setting window bounds for windows which are not user-modifiable


inolen:
I regularly use nuklear to render fullscreen elements like so:

https://cloud.githubusercontent.com/assets/460296/20913584/fbd476f6-bb2d-11e6-9068-54ef177a2a49.png

However, due to the way windows are cached, I must call `nk_window_set_bounds` after each `nk_begin` to force the bounds for these fullscreen elements or else the bounds remain what they were initially created with, which breaks down as the window itself is resized.

Would it make any sense for windows which are _not_ user-modifiable (e.g. not flagged with without `NK_WINDOW_MOVABLE` / `NK_WINDOW_SCALABLE`) to always have their bounds set by `nk_begin`, or is that an unwanted inconsistency in the API?

vurtun:
Hey, yeah this is one of my API crimes I committed with nuklear. Basically I should have used either a binding style API or even better a bindless API with IDs to modify a window directly and remove them from my immediate mode API. Since I don't want to rewrite the whole API your solution seems enticing. The problem like you said is inconsistency, the question is to what extend. Is this behavior really inconsistent because basically if the window cannot change by user input it should be able to be modified directly. It is almost something to be expected, but it is just not really user friendly.

Another question is if nuklear should be kept consistent and I would only provide a hack which to include if you need it or if it should be directly included into nuklear. But since this an actual bug that can cause problems I think it is worth it to make these changes directly. If you want then you can create a PR or if you don't have the time then I will make these changes. The changes are not really complex just add an `if` inside `nk_begin_titled` to set the window position,size if window is not scalable and movable.

dumblob:
>Basically I should have used either a binding style API or even better a bindless API with IDs to modify a window directly and remove them from my immediate mode API.

Well, I was thinking about a "bindless"-like API variant some time ago and decided to give it a try for our needs. @vurtun, do you have something particular in mind we should follow or take into account? So far we're designing the "new" API, but because issues related to windowing are a bit repetitive requests, I'd like to rather first discuss the new API with you for the case you would like to incorporate such a redesigned API to Nuklear upstream instead of "forking" Nuklear. We're prepared to totally redesign Nuklear API if needed.

inolen:
@vurtun if you have the bandwidth to make the changes, that'd be great, otherwise I'll submit a pr in a week or two.

vurtun:
@inolen thanks for the PR.
[...]

PR #309, opened 2016-12-29
Set window bounds in nk_begin when the window isn't user-resizable

inolen:
As mentioned in https://github.com/vurtun/nuklear/issues/305


vurtun:
Thanks for the PR looks good!

PR #313, opened 2017-01-01
build fix for gcc

inolen:
GCC on my travis.ci builder was complaining about the parens in this expression.

I believe the intent of the expression was to check if the window was `minimized && !closed`.


vurtun:
Yeah stupid mistake on my side, thanks for PR.

PR #315, opened 2017-01-05
Allegro5 nuklear backend and touchscreen support

seibelj:
I have implemented a new nuklear backend for Allegro v5 (http://liballeg.org/). I have included a Readme file for more information.

I tweaked the nuklear.h library slightly. I added a new  optional define called `NK_INCLUDE_DYNAMIC_SOFT_KEYBOARD`. The only thing this does is call `open_keyboard()` and `close_keybord()` functions on the context (if they are implemented) when text entry begins and ends. This is necessary to allow the programmer to know when to show and hide the soft keyboard on a touch screen device.

I also added a new `NK_KEY_TEXT_SELECT_ALL` command for `nk_input_key()` which seemed to be an oversight, as the function was already implemented in `nuklear.h`.

The two Objective-C  (`KeyboardHandleriOS.h` and `KeyboardHandleriOS.m`) files are included for reference purposes for iOS developers, as they took many hours to figure out and will be very useful to have available. They are also described in the Readme.

I have successfully used nuklear on Android phones, Android tablets, iPhones, and iPads. It is a very nice experience as a developer.


vurtun:
Thanks for the PR there is some good stuff in there. This will probably take some time to review. But  I am not sure about the `open_keyboard\close_keybord` callbacks. Basically if you add a callbacks to this library really think about if there any alternative.

For example correct me if I am wrong but I already provide a way to open and close a screen keyboard. Every edit box widget returns flag `NK_EDIT_ACTIVATED` if a edit widgets was activated and `NK_EDIT_DEACTIVATED` if deactivated. So you basically just do:

```c
nk_flags flags = nk_edit_string(ctx, ...);
if (flags & NK_EDIT_ACTIVATED)
    /* open screen keyboard */
if (flags &  NK_EDIT_DEACTIVATED)
    /* close screen keyboard */
```

to achieve the same effect as using callbacks with more flow control on the user side (You could write a wrapper around `nk_edit_xxx` for touch which does this automatically). 
However I agree that `nk_property` is missing the same flag (at least in the API) so it would make sense to add additional `nk_property` functions to add these flags. (I also dislike passing the `nk_context` struct, which is not required with flags, to any `do` widget functions. I basically try to keep the context related functions separated from their widget implementation).
As for select all text makes perfect sense seems like I forgot to add it.

Overall a good PR request even if I preferred if it was split into multiple parts to make it easier to review, accept or deny parts. By that I mean I am forced to refuse this entire PR in it's current form even if there are big parts I would pull without question.

seibelj:
Thanks so much for your time. I removed all of the keyboard callback code after verifying that `nk_flags` mechanism you proposed worked. That is much more elegant, thank you.

The PR shows changes to `nuklearh.h` that you have already made, I'm not sure why it thinks I'm making them. I could reopen the PR if you would like, that might look cleaner. The only changes I am now making to `nuklear.h` are adding the `NK_KEY_TEXT_SELECT_ALL` command and changing a variable name from `nil` to `none` because `nil` is a reserved keyword in Objective-C.

Now it's really just the new allegro5 backend and the associated code, which should make the risk much smaller despite being a big PR.

PR #316, opened 2017-01-06
GDI+ images support

DeXP:
There are 2 user functions added: `nk_gdip_load_image_from_file` and `nk_gdip_load_image_from_memory`. The second one requires function from `ole32.lib`


vurtun:
Thanks for the PR looks good.

PR #327, opened 2017-01-20
GDI+: loading TTF font from file

DeXP:
Also font API rename to be the same with images.


vurtun:
Thanks for the PR. I really hate to pull this since it is a breaking change to an API. But since this is not the actual library I guess I will accept it. It was my mistake in the first place to not make sure in the beginning that all names fit the same style so I guess I am to blame for this. Sorry everyone for having to change the API!

PR #330, opened 2017-01-23
Remove warning: 'nk_to_upper' defined but not used

DeXP:
There is function `NK_INTERN int nk_to_upper(int c)` in `UTIL` section. Implementation is almost the same to `NK_INTERN int nk_to_lower(int c)`. But only `nk_to_lower` used in Nuklear's code. To remove this warning we can: 
1) Comment i.e. remove `nk_to_upper` from code. But it can break someone's code, if he used it in module with `NK_IMPLEMENTATION` defined.
2) Hide it under define. Same. 
3) Just use it somewhere in code. 
I replaced two calls of `nk_to_lower` to `nk_to_upper`. It removes the warning, code works the same. The bad side - complication of the code. "Why `nk_to_upper` is used here, but not `nk_to_lower` as one line up?"
Maybe there exists better solution? Or just add the comment about this line to code?

### DIFF
--- a/demo/x11_opengl2/nuklear_xlib_gl2.h
+++ b/demo/x11_opengl2/nuklear_xlib_gl2.h
@@ -10,8 +10,8 @@
  *
  * ===============================================================
  */
-#ifndef NK_XLIB_GL3_H_
-#define NK_XLIB_GL3_H_
+#ifndef NK_XLIB_GL2_H_
+#define NK_XLIB_GL2_H_
 
 #include <X11/Xlib.h>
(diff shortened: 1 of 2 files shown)


vurtun:
While the warning is luckily not critical it is annoying. I like your solution of just using 'nk_to_lower' and having both 'nk_to_upper' and 'nk_to_lower'. Since they are mostly used in cases where upper or lower case does not matter it shouldn't really confuse anyone.

DeXP:
Ok, I will split my future commits into different PR's by theme. Thanks for merging, this warning was really annoying.

PR #334, opened 2017-01-30
added CMake build for Linux

MattGuerrette:
Due to the growing use of CMake as a build system and the use of CLion as a development environment I've created a cmake build for nuklear. Currently, this only supports Linux but it could be extended to support building on Windows using CMake as well.

One thing to note however, is that currently there isn't a choice of which example to build. CMake will attempt to generate Makefiles for every example which means that **ALL** dependencies are required to be installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,86 @@
+cmake_minimum_required(VERSION 3.0)
+project(Nuklear)
+
+if(UNIX)
+    include_directories("${CMAKE_SOURCE_DIR}")
+
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(GLEW REQUIRED glew)
+    pkg_check_modules(GLFW REQUIRED glfw3)
+    pkg_check_modules(SDL2 REQUIRED sdl2)
+    pkg_check_modules(OpenGL REQUIRED gl)
+    pkg_check_modules(X11 REQUIRED x11)
+    pkg_check_modules(ALLEGRO REQUIRED allegro-5
+                                       allegro_image-5
+                                       allegro_primitives-5
+                                       allegro_main-5
+                                       allegro_ttf-5
+                                       allegro_font-5)
+
+    file(GLOB ALLEGRO5_SOURCE ${CMAKE_SOURCE_DIR}/demo/allegro5/*.c)
+    add_executable(allegro5 ${ALLEGRO5_SOURCE})
+    set_target_properties(allegro5 PROPERTIES
+            RUNTIME_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}/demo/allegro5/bin/)
+
+    file(GLOB GLFW_OPENGL2_SOURCE ${CMAKE_SOURCE_DIR}/demo/glfw_opengl2/*.c)
+    add_executable(glfw_opengl2 ${GLFW_OPENGL2_SOURCE})
+    set_target_properties(glfw_opengl2 PROPERTIES
+            RUNTIME_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}/demo/glfw_opengl2/bin/)
+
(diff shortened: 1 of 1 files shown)


dumblob:
It's weird, that the multiplatform `Makefile`s are not enough for CLion and systems using CMake. Why should we introduce another dependency (CMake is quite bloated)? Is it really just because of CLion lacking enough support for such tiny plain `Makefile`s?

I'm unable to find any good reason for maintaining the CMake builds. What am I missing?

MattGuerrette:
I think your are mincing words. bloated is not the same as powerful. Maintaining a CMake build would allow for integrating dependencies directly into the project build for platforms that don't have a package manager (eg Windows). Almost all of the dependencies Nuclear links use CMake and can therefore be added as submodules and integrated.

Open source projects should really try leveraging these tools rather than operate on dated Makefile systems . Many of them already do and there is no reason why Nuklear can't also.

dumblob:
@MattGuerrette please apologize me as the following might sound rude, but that's what popped in my mind.

>I think your are mincing words. bloated is not the same as powerful.

There is absolutely no need to use something *powerful* for the tiny job the already existing and fully multiplatform and standardized Makefiles do.

>Maintaining a CMake build would allow for integrating dependencies directly into the project build for platforms that don't have a package manager (eg Windows).

1. Each project uses it's own build system (if any - think of e.g. [Ekam](https://github.com/sandstorm-io/ekam )). CMake used to be slightly more used on Windows, but nowadays not any more that much. Microsoft open-sourced the widespread Visual Studio (which supports `make` out of the box) and other tools, JavaScript based build systems are on rapid rise, etc. With the advent of by-default built-in support for unix/Linux stuff in Windows, there is no more something like "build system which rules majority of projects" (there are piles of discussions about build systems and the amount of their usage in multiplatform projects - in the end one realizes it's useless to support and maintain something else than the absolute multiplatform minimum like a plain `Makefile`).

1. Integrating the Nuklear dependency into any CMake project is a matter of about 3 lines. I would definitely recommend to put these 3 lines to some **Best practises** section in this repository README file instead of introducing this useless CMake build file in the repository.

>Almost all of the dependencies Nuclear links use CMake and can therefore be added as submodules and integrated.

I'm afraid it's an utterly wrong idea to add these **examples and demos** as submodules to any project and integrating them. They're not designed for that and will most probably not be in the near future. They're meant really just to give you a very precise overview how to use few common backends. Nothing else.

If we wanted to use the Nuklear repo e.g. as a standalone git subrepository, we would need to change quite a lot of things in the Nuklear repo to make it safe and cross-project usable without user intervention. Is there any written standard or at least profound tool which "combines" projects of this kind (i.e. a library written in a compiled language) into a larger project?

>Open source projects should really try leveraging these tools rather than operate on dated Makefile systems .

Good point and I totally agree.

>Many of them already do and there is no reason why Nuklear can't also.

There is a very firm reason - staying KISS. If `Makefile` works everywhere (and without introducing any dependency), offers the same level of maintenance complexity and in this case offers exactly the same user friendliness (readability, possibility of processing automation, etc.), it doesn't make absolutely any sense to add something else.

paulsapps:
I think CMake is the standard for building open source software on Windows - while it might seem too much for nuklear don't forget that nuklear will be yet another dependency in bigger projects therefore it does make sense. 

Also cmake can generate regular make files if need be. VS really isn't designed for make files at all, its sln/vcproj files are the standard which cmake can generate. I've never had a good time using VS and make files.

vurtun:
Thanks MattGuerrette , @dumblob and @paulsapps. So my personal technical opinion of CMake is that it is absolutely overly complex and bloated. However this is not about my personal opinion, but rather my opinion as a maintainer. CMake is widely adopted and almost somewhat expected for windows. So I am willing to support CMake. At least If some conditions are met. First of all if I will pull CMake into nuklear it has to be mutliplatform with at least windows, linux and os support. Second it does not make sense to have one big cmake file since no one should, needs or has all dependencies to be installed (especially under windows). So I would allow one CMake file for each demo provided it does support windows, linux and os.

Issue #333, opened 2017-01-23
OpenGL big font rendering cuts the edge


DeXP:
![wordgame-big-font](https://cloud.githubusercontent.com/assets/4019495/22210711/c8e6d5ce-e19b-11e6-9039-3c8bfbc1df1a.png)
`nk_font_atlas_add_from_file(atlas, "DejaVuSans.ttf", 60, 0);`
"P" letter is trimmed on the right; "o" from right and bottom etc.
It's for all fonts, not just DejaVuSans. Extremely seen on Sans, not Serif fonts.
Maybe antialiasing problem and/or texture size mismatch?
Front end: X11, SDL, GLFW. OpenGL both 2 and 3.

vurtun:
Interesting problem. This could have a number of different causes going from harmless to god aweful.  Possible sources could drawing inside `nk_draw_list_add_text` and 'nk_draw_list_push_rect_uv'. Then font baking and glyph packing inside `nk_font_atlas`  in function `nk_font_atlas_bake`. The absolute worst case would be inside font rasterizer.

wolfgolang:
9449  for (i=0; i < num_rects; ++i) {
                rects[i].h = rects[i].h+2;      //add this 
 9450      struct nk_rp__findresult fr = nk_rp__skyline_

PR #339, opened 2017-02-09
Fix panel clipping

DeXP:
`layout->bounds` cannot be used (loc 16615) since we modify it up here:
`layout->bounds.x += panel_padding.x;`  (16471)
`layout->bounds.w -= 2*panel_padding.x;` (16472)
Because of it there was no ability to `nk_draw_image` attached to window's edge. There alway was empty lines of right and left.

This code breaks the other's code. Popups, for example. My paddings are 10px. To show popup at (20,20) without this patch I must use `nk_rect(10, 20, ...`. With this patch I must use `nk_rect(20, 20, ...`, not depending on paddings. I think (20,20) is a correct constant for (20,20) coordinate.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -16613,7 +16613,7 @@ nk_panel_begin(struct nk_context *ctx, const char *title, enum nk_panel_type pan
 
     /* set clipping rectangle */
     {struct nk_rect clip;
-    layout->clip = layout->bounds;
+    layout->clip = win->bounds;
     nk_unify(&clip, &win->buffer.clip, layout->clip.x, layout->clip.y,
         layout->clip.x + layout->clip.w, layout->clip.y + layout->clip.h);
     nk_push_scissor(out, clip);


DeXP:
Also, can you give me some hints about my [OpenGL big font](https://github.com/vurtun/nuklear/issues/333) issue? Where its better start to dig? I will try to fix it by myself too.

vurtun:
I am not sure this is correct. If you want to have a window at position (20,20) you need to set padding to zero, not remove padding all together. Your version does not work for windows with header but without menu (check skinning.c for example).

DeXP:
ok, agreed. Changing `layout->bounds` to `win->bounds` is not correct. Then the question is about this lines:
```c
layout->bounds.x += panel_padding.x; (16471)
layout->bounds.w -= 2*panel_padding.x; (16472)
```
I want to draw on my window's canvas via `nk_draw_image`. The image is bigger than window. In result I want to get something like this:
wordlase-mainmenu
Now to draw it correctly I need:
```c
tmp = ctx->style.window.padding;
ctx->style.window.padding = nk_vec2(10,20);
nk_draw_image(...);
ctx->style.window.padding = tmp;
```
Why window's background picture must depend on paddings? Why only left and right depends on it, but not top and bottom? I think, this logic is not correct too.

vurtun:
To clarify do you want to draw the image to an OS window or nuklear window? If you want to draw an image to a nuklear window then temporarily set `ctx.style.window.fixed_background` to your image.

DeXP:
Nuklear window. And not only one, but a bunch of images. Coordinates calculated dynamically.

vurtun:
So it would probably make sense to use `nk_image` or `nk_subimage` instead. The problem seems to me that you are using `nk_draw_image` which is a low level drawing API call which does not care for `ctx->window.padding` at all only clipping (which is probably your problem). What I would do is, if I understand you correctly, is to temporary set `ctx->window.padding = nk_vec2(0,0)` and then use `nk_space_begin`, `nk_space_push`, `nk_space_end` to directly place your images to screen.

DeXP:
Yep, `nk_space_push` with `nk_image` draws my background correctly. But it moves the layout cursor too. I can't place buttons or any other form elements on top of that...

I'm making game on pure Nuklear (without direct OpenGL usage): [Wordlase](http://steamcommunity.com/sharedfiles/filedetails/?id=861528240). Every screen is a nuklear window, stretched to fullscreen.

vurtun:
Definitely not a perfect solution. You basically have to temporarily store `at_x` and `at_y` attribute from `nk_panel`, draw your image with `nk_space_push` and `nk_image` and load both back. Not a nice way of doing it. You could probably use one of the stack function to temporarily push and pop the value but still ugly. At the moment I can't really think of any good solution for this.

DeXP:
Hm. Are you sure about `at_x` and `at_y`?
````c
if( nk_begin(...) ){
    struct nk_panel* layout = nk_window_get_panel(ctx);
    float my_at_x = layout->at_x;
    float my_at_y = layout->at_y;
    nk_layout_space_begin(ctx, NK_STATIC, totalHeight/3, n);
    {
        ...
        nk_layout_space_push(ctx, nk_rect(x, y, width, height));
        nk_image(ctx, img);
    }
    nk_layout_space_end(ctx);
    layout->at_x = my_at_x;
    layout->at_y = my_at_y;
    ...
    /* form code */
}
````
My widgets still starts from `totalHeight/3`...

**Comment by vurtun** [comment]
Ah shit you are right. While it is true that `at_x` and `at_y` control widget placement it is `panel->row.height` that control row progression. So you only need to set it to zero after your `nk_layout_space_end` call. So basically:

```c
if( nk_begin(...) ){
    struct nk_panel* layout = nk_window_get_panel(ctx);
    float my_at_x = layout->at_x;
    float my_at_y = layout->at_y;
    nk_layout_space_begin(ctx, NK_STATIC, totalHeight/3, n);
    {
        ...
        nk_layout_space_push(ctx, nk_rect(x, y, width, height));
        nk_image(ctx, img);
    }
    nk_layout_space_end(ctx);
    layout->row.height = 0;
    ...
    /* form code */
}
```

DeXP:
Yes! Now it works! Great thanks.
I still need `ctx->style.window.padding = nk_vec2(0,0)` before `nk_begin`, but it works as I wanted.

vurtun:
About padding. If you don't need a scrollbar it should be possible to directly set padding back to its old value and it should work:

```c
nk_style_push_vec2(ctx, &ctx->style.window.padding, nk_vec2(0,0));
if( nk_begin(...) ){
    struct nk_panel* layout = nk_window_get_panel(ctx);
    nk_layout_space_begin(ctx, NK_STATIC, totalHeight/3, n);
    {
        ...
        nk_layout_space_push(ctx, nk_rect(x, y, width, height));
        nk_image(ctx, img);
    }
    nk_layout_space_end(ctx);
    nk_style_pop_vec2(ctx);
    layout->row.height = 0;
    ...
    /* form code */
}
```

DeXP:
It doesn't really matters, save padding to temporary variable or use `nk_style_push_vec2`. I must do it before `nk_begin`, not in my background-render function...

PR #344, opened 2017-02-11
Make NK_API 'static inline' for NK_PRIVATE version

anton-bondarev:
When NK_PRIVATE is defined, interface routines become static in a header, so it's better to use inline declarations

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -272,7 +272,7 @@ extern "C" {
  */
 #ifndef NK_API
 #ifdef NK_PRIVATE
-#define NK_API static
+#define NK_API static inline
 #else
 #define NK_API extern
 #endif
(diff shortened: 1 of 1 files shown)


vurtun:
This library uses C89 by default which does not have an `inline` keyword. So it would make sense to only define it if checked for C99 or C++. So something like this:

```c
 #ifndef NK_API
  #ifdef NK_PRIVATE
    #if (defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199409L))
      #define NK_API static inline
    #elif defined(__cplusplus)
      #define NK_API static inline
    #else
      #define NK_API static
    #endif
  #else
    #define NK_API extern
  #endif
```

anton-bondarev:
Yes, you're right. I did as you suggested.

Issue #352, opened 2017-02-25
NK_COMMAND_CUSTOM


ClaasJG:
Good evening,

It is easy to extend this library (#265 as example) as long as you are working with nuklear.h but pretty hard when using a binding to use the provided functionality in an other language.

A custom command type would be useful, one that simply calls a callback that can be used to populate the  draw_list.

I made a small 'implementation mockup' here: https://github.com/ClaasJG/nuklear/commit/4c25cd34f5ef83b2bdf403ec332fdd159347e332

Have a nice day,
ClaasJG

vurtun:
No problem. Just create a PR. I see no reason to not add a custom command type. It does not break any existing code or prevents someone to just add his own command version in C.

PR #353, opened 2017-02-25
pull request for #352

ClaasJG:
pull request for #352 providing a custom command type which calls a callback to fill the draw_list


vurtun:
I had to do some changes to make it fit. I changed the `nk_draw_lilst` to `void*` since it is an optional struct. Your version will still work however. I also don't want a `nk_custom` widget since it is quite a niche I would prefer if the user creates it.

ClaasJG:
I added `nk_custom` just for consistency, I guess it is not really necessary.
Thank you for an awesome library with great support!

Have a nice evening
-ClaasJG

PR #389, opened 2017-03-30
Several fixes

rokups:
Fixed examples using `nk_input_scroll()` api properly.

~~Fixed memory leak in `nk_draw_list_path_line_to()` - points were allocated but not really used.~~ Well looks like it had side-effects. AddressSanitizer still reports it as a memory leak though..

Initialized `ws` variable to 0 in `nk_scrollbar_behavior()`. Uninitialized variable was being passed to `nk_button_behavior()` which performed `nk_widget_state_reset(state)0;` on it which expands to if/else.

So thats that, some low-hanging fruits pointed out by AddressSanitizer.

### DIFF
--- a/example/canvas.c
+++ b/example/canvas.c
@@ -307,7 +307,7 @@ static void error_callback(int e, const char *d){printf("Error %d: %s\n", e, d);
 static void text_input(GLFWwindow *win, unsigned int codepoint)
 {nk_input_unicode((struct nk_context*)glfwGetWindowUserPointer(win), codepoint);}
 static void scroll_input(GLFWwindow *win, double _, double yoff)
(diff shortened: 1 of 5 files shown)


GreenLightning:
`scroll_input` is used as GLFW [scroll callback](http://www.glfw.org/docs/latest/input_guide.html#scrolling) and also receives the x-offset, so the code should probably look like this:

```c
static void scroll_input(GLFWwindow *win, double x_offset, double y_offset)
{nk_input_scroll((struct nk_context*)glfwGetWindowUserPointer(win), nk_vec2((float)x_offset, (float)y_offset));}
```

vurtun:
Thanks for the PR I forgot to transition all examples.

rokups:
Any reason why PR was not accepted?

vurtun:
ops wrong button.

PR #422, opened 2017-04-30
Add software rasterizer library and demo.

siro20:
Add software rasterizer library and demo.
The software rasterizer is to be used with raw framebuffer
devices, where no GPU or X11 is available.
The demo emulates a raw framebuffer on X11 using XShmImage / XImage.
Fonts are rendered software only using nearest filtering.

Tested:
The library has been tested on Lenovo Thinkpad T500 and is able to render more than 30fps on a single core with no further optimizations and VSNYC enabled.

TODO:
Improve font rendering by using filters.
Window resizing isn't supported by the X11 demo application.

Signed-off-by: Patrick Rudolph <siro@das-labor.org>

### DIFF
--- a/demo/x11_rawfb/Makefile
+++ b/demo/x11_rawfb/Makefile
@@ -1,0 +1,13 @@
+# Install
+BIN = zahnrad
+
+# Flags
+CFLAGS = -std=c89 -pedantic -O2 -Wunused -DRAWFB_XRGB_8888
+
+SRC = main.c
+OBJ = $(SRC:.c=.o)
+
+$(BIN):
+	@mkdir -p bin
+	rm -f bin/$(BIN) $(OBJS)
+	$(CC) $(SRC) $(CFLAGS) -D_GNU_SOURCE -D_POSIX_C_SOURCE=200809L -o bin/$(BIN) -lX11 -lXext -lm
--- a/demo/x11_rawfb/main.c
+++ b/demo/x11_rawfb/main.c
@@ -1,0 +1,221 @@
+/* nuklear - v1.17 - public domain */
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdarg.h>
+#include <string.h>
+#include <limits.h>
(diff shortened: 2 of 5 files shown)


dumblob:
Thanks for splitting it into two commits.

I finally took a closer look and it seems the code needs some love. First I tried to track down why the dimension of the buffer can't be changed in runtime dynamically and didn't find any obstacles - it would require just decoupling from the initialization functions and otherwise everything shall work. So I definitely agree with the TODO item `Window resizing isn't supported by the X11 demo application`.

Second thing is e.g. the `nk_rawfb_render()` function using a global variable. Globals definitely shouldn't be anywhere (I didn't check for other occurrences, but I'll do it once the code settles down a bit) and instead usual function arguments should be the way to go. Also it would be nice to mark API functions with information about thread safety (nearly everything seemed safe, but again, I didn't do any deep analysis).

siro20:
I've made it thread safe by using a context instead a static object.
I fixed the dynamic resize.
Added a comment about thread safety.

dumblob:
@siro20 thanks.

I'm currently out of time and will get back to this first in about two months :cry:. In the mean time anybody can do a review to open the path to the mainline :wink:.

vurtun:
Hey @siro20,
I went over, read and run your code. Looks really nice I like it. However I noticed a few issues. Could you please change the coding style to match closer to nuklear ( `{`  always on same line except for functions and `if` with only one expression don't have any `{}` at all). Second and more important issue is a clipping bug I found. If you run the `overview.c` UI and scroll, the clipping rect seems to be off. Especially visible  for  input boxes:

clipping

siro20:
I fixed the following points:
* Use nuklear's coding style
* Fixed clipping bug

vurtun:
Style looks OK, there are still some style issue by I can change these by myself. However while the lower window clipping rectangle seems to work now the upper clipping bounds seems to be bugged:

bug

siro20:
Fixed upper scissors
Added support for text scissors

PR #436, opened 2017-05-17
double click support in demo code for Windows platforms

mmozeiko:


vurtun:
Thank you for your work and PR. Was expecting more hassle but seemed to be surprisingly simple solution.

PR #450, opened 2017-06-04
Implemented nk_create_image and nk_delete_image for the win32 gdi

Fra-Ktus:
Implemented nk_create_image and nk_delete_image.
the image is simple raw buffers in 24 bits (no alpha), in my case they are buffers received over the network.
This patch allows to create an interface using nk_images and nuklear controls together...


vurtun:
Thanks for the PR. Since I don't have a windows machine I cannot really test it so I have to trust you on this one.

PR #475, opened 2017-07-06
Remove unnecessary semicolons

valera-rozuvan:
GCC (version 7.1.0) reports these semicolons as unnecessary (using `-pedantic` flag), and compiles fine without them. We should remove them.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -11165,11 +11165,11 @@ struct nk_font_baker {
     struct nk_tt_pack_range *ranges;
 };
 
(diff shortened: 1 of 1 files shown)


vurtun:
Hey, close but no. The problem lies in https://github.com/vurtun/nuklear/blob/master/nuklear.h#L4234. If you want to update the PR just remove that semicolon, if not I will push a commit myself.

valera-rozuvan:
I'll just close it.

PR #492, opened 2017-08-11
Optimize arc drawing for fewer sin/cos calls

ettoretorti:
Change nk_draw_list_path_arc_to so that it calls sin and cos 4 times
instead of (segments * 2) times. This is achieved through the use of
the trigonometric angle addition identities.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -8586,11 +8586,39 @@ nk_draw_list_path_arc_to(struct nk_draw_list *list, struct nk_vec2 center,
     NK_ASSERT(list);
     if (!list) return;
     if (radius == 0.0f) return;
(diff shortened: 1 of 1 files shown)


ettoretorti:
Obviously the gain depends on how expensive it is to call sin/cos. This patch uses fewer additions/multiplications per iteration than using the provided nk_cos/nk_sin procedures.

I think it's safe to say that the stdlib sin/cos procedures are significantly more expensive, which results in this patch providing an even greater performance improvement if those are used.

vurtun:
Thanks for the PR, great work. 
I changed it up a little bit. I converted the C++ comments into C89 comments to make it fit and added a reference to the addition and subtraction theorems.

PR #494, opened 2017-08-13
SDL software rendered

slvtmn:
SDL software rendered 
Dependence on TTF_Font.h and SDL.h


mmozeiko:
I don't think this its accurate to call this "software rendered". SDL_Renderer is GPU backed renderer. It uses OpenGL, D3D or similar API to do actual rendering. The documentation is pretty clear on this: https://wiki.libsdl.org/CategoryRender

vurtun:
In addition to what @mmozeiko said. I also noticed that a number of primitives cannot be drawn directly by SDL and the coding style is completely off. So I am a little bit hesitant to add this demo backend. I now have three SDL demo backends which should be enough.

hbiblia:
Seria muy bueno que nuklear tenga soporte a SDL2 con SDL_Rendered

hbiblia:
Estoy trabajando en Nuklear SDL_Native - https://github.com/hbiblia/nuklear/tree/master/demo/sdl_native

hbiblia:
Yo también estoy trabajando en algo similar - [SDL2_Native](https://github.com/hbiblia/nuklear/tree/master/demo/sdl_native)

PR #496, opened 2017-08-15
Fix typos in comments

ettoretorti:
This commit fixes all the typos I could find in the comments by running them through a spellchecker.


vurtun:
Thanks for the PR.

PR #512, opened 2017-09-04
XLib: stb_image support with bit transparency

DeXP:
X11 do not support transparent images itself. So the clip mask was calculated in this code. X11 ClipMask support only bit transparency.
Another negative side is a bad performace on big transparent images. But it work fast with opaque
 images.

Original OpenGL render:
xlib-opengl

Xlib without transparency:
xlib-no-alpha

And final Xlib variant with bit transparency:
xlib-alpha

It's not good enough to use in games, semi-transparent images. But pretty good for UI icons, I think.


DeXP:
Useful links about transparent X11 images:

- [Puppy In-House Development](http://www.murga-linux.com/puppy/viewtopic.php?t=89272&start=390&sid=f803cf50591c039770f3acd0328040e5)
- [How to make part of an XImage transparent ?](https://groups.google.com/forum/#!topic/comp.windows.x/T9MjoX6V7Ag)
- [Using transparent images](http://www.dis.uniroma1.it/~liberato/screensaver/trans.html)

vurtun:
Hey,
thanks for the PR. While I absolutely like the changes I think it would be best to introduce an additional configuration define to make it optional. My main reasoning would be that each of the `nuklear_xxx` should be as self contained as possible. While I provide a version of `stb_image.h` with this repository I don't want to directly depend on it.
So it would be great to have a define like `NK_XLIB_INCLUDE_STB_IMAGE` and  `NK_XLIB_IMPLEMENT_STB_IMAGE` (or similar naming). The reason for two defines is that if a user already implements stb_image.h it would cause errors.  
`

DeXP:
I added both `NK_XLIB_INCLUDE_STB_IMAGE` and `NK_XLIB_IMPLEMENT_STB_IMAGE` to the code.

PR #522, opened 2017-09-21
missing parameter to nk_window_set_bounds call

valera-rozuvan:
`nk_window_set_bounds` function expects 3 parameters, the 2nd of which is the window `name`


vurtun:
Sorry noticed to late that you already created a PR fixing the issue and pushed commit 0b95e3e5a018acd839efc9386dccb73ee67c0a2f. Regardless the issue should be fixed now.

Issue #374, opened 2017-03-17
No DOC at all?


RUSshy:
Hello

I can't see where is the doc, also no wiki to explain and list each widget and how to use them like text input for example ?

Thanks

RUSshy:
I'll start few pages on wiki for the controls i think i know how to use them, but my english is not very good (russian)

RUSshy:
I created the sidebar here: https://github.com/vurtun/nuklear/wiki

I think the layout is fine, now need to edit/create each page xD

RUSshy:
For consitensy with the screenshots, here is my settings:


Window Size: 800 x 600
Window Title: Demo

Nuklear Bounds: WINDOW_WIDTH/2 - 110, WINDOW_HEIGHT/2 - 110, 220, 220)
Nuklear Window Name: Nuklear

Background: 28, 48, 62

cxong:
This is a good initiative - this project really needs an API reference - but I wonder if it's better to use something like doxygen hosted on readthedocs.org?

ricanteja:
I think @RUSshy has the right idea with the Wiki but I believe it should be reserved for tutorials. For in-depth line by line descriptions of what each structure and function does we should be looking towards doxygen.

vurtun:
OK it seems that documentation is currently one of the biggest problem with `nuklear`. So far the only reason I did not write more documentation was not lazyness. Instead I always liked how all functions declarations just read like a reference. But I have seen enough complains about it that I think it is time to hold a vote. So I am willing to write documentation. In the beginning I would start with documentation for functions declaration. In fact over a year ago I already documented *every* [functions declaration](https://github.com/vurtun/nuklear/blob/daae1b880434c5f331c894b0dc58d73650462a5a/zahnrad.h).
I liked that particular style of first having a list of all functions as reference and then descriptions of every functions. 

So here are the two options:
1. Simple non-descriptive list of functions without documentation
2. Documentation for every function like I had in the past

dumblob:
>1. Simple non-descriptive list of functions without documentation

This doesn't make any sense as e.g. [Doxygen](http://www.doxygen.org/index.html ) (the most famous doc tool for many programming languages) does this by default.

>2. Documentation for every function like I had in the past

Yep, this sounds to be the way to go. Just make sure Doxygen can read **all** the descriptions and you're fine forever :wink:.

vurtun:
[...]

PR #523, opened 2017-09-21
Docs

cosier:
After a bit of discussion over at #374 , we came to a few conclusions...

Here we have a basic setup for documentation generation.
This pull request uses doxygen + sphinx to output static html into docs/build/html

We then overlay another repository into docs/build/html which serves as Github pages repository.

Which is the reason I setup org nuklear-ui, so we may use https://nuklear-ui.github.io as the documentation url root.

_(I have added @vurtun as owner to that org, if that is okay with you)_

-------------

Steps to generate documentation

1. `$ docs/setup_docs.sh` To check and document any needed dependencies we are using (ie. plugins)

2. `$ docs/make_docs.sh` Runs doxygen + sphinx generation

3. `$ docs/release_docs.sh` Auto commit and push to github pages repo

With this in place, we can move onto making the docs great and adding more pages.
Open to any other ideas or discussion.


valera-rozuvan:
Way to go!

cosier:
Currently, we the sphinx html docs are available online for reading at
https://nuklear-ui.github.io/

vurtun:
At least if I run `$ docs/setup_docs.sh ` I get an error under arch linux:

```
Command "/usr/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-s2ujssic/MarkupSafe/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-6l9evoeq-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-build-s2ujssic/MarkupSafe/
```

PR #529, opened 2017-10-01
GDI+ multiple fonts support

DeXP:
Each custom loaded font requires own `GpFontCollection`. So there is array of font collections now.


vurtun:
In this [line](https://github.com/vurtun/nuklear/pull/529/files#diff-7141b2e4ef61047b5d0175f537069540R647) the preprocessor define . Is this correct. Seems odd for me to check for `0==1`. Also could you please prefix `getCurFontCollection` with `nk_` to `nk_gdip_getCurFontCollection`. Namespaces are really important in single header library context. Other than that looks good.

DeXP:
Fixed

PR #538, opened 2017-10-17
Improvement to SFML GL2 demo code

metiscus:
Adds protection of the texture matrix stack which allows the user to use sf::Sprites with this code unmodified otherwise.

Using sf::Sprites with this example code, results in the text rendering as blurred boxes and the window panels being extremely transparent. Protecting the texture stack prevents this so if someone copy pastes this code they are not frustrated.

### DIFF
--- a/demo/sfml_opengl2/nuklear_sfml_gl2.h
+++ b/demo/sfml_opengl2/nuklear_sfml_gl2.h
@@ -81,6 +81,9 @@ nk_sfml_render(enum nk_anti_aliasing AA)
     glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
 
     glViewport(0, 0, (GLsizei)window_width, (GLsizei)window_height);
+    glMatrixMode(GL_TEXTURE);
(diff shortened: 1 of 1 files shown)


dumblob:
@metiscus Could you maybe add this comment about this protection matrix to the code itself? If I read the code currently I can't answer my question why the heck is the code setting the `GL_TEXTURE` matrix to identity.

metiscus:
The issue is that SFML appears to modify the GL_TEXTURE stack while this code works with the implicit assumption that the texture stack has an identity matrix loaded. If you run this code with SFML in a scene in which you do not draw any texture based sprites (possibly if you only draw texture based sprites that are 1:1 with the underlying texture / image) then the GUI will render properly because SFML does not modify the texture stack in this case. If, however, you render a sf::Sprite (potentially that you have used a non 1:1 mapping to the texture/image on) then when this code draws the texture stack is not at identity so the transformation is applied to fonts and underlay texture so the UI is corrupted.

I am not sure how you would want that to be written up but I leave it to you to wordsmith and stick it wherever.

PR #549, opened 2017-11-11
Fix D3D9 rendering in certain scenarios

GreenByteSoftware:
This fixes D3D9 rendering in certain scenarios, like drawing in video games on top of them. Here is a screenshot of the issue that had been occurring:

Image of the issue

The black part was supposed to be a window. The window would render fine for one or a couple of frames and then turn black. This fix makes the rendering similar to what ImGui does in their render function by recreating the D3D9 state on every frame and releasing it after.

### DIFF
--- a/demo/d3d9/nuklear_d3d9.h
+++ b/demo/d3d9/nuklear_d3d9.h
@@ -68,9 +68,53 @@ static struct {
 } d3d9;
 
 NK_API void
+nk_d3d9_create_state()
+{
+    HRESULT hr;
+
+    hr = IDirect3DDevice9_BeginStateBlock(d3d9.device);
+    NK_ASSERT(SUCCEEDED(hr));
+
+    /* vertex format */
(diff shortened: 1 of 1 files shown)


mmozeiko:
@vurtun, I would like to suggest to revert this commit. It messes up a lot of things - state block is now created every frame. The purpose of it is to encapsulate all needed state, so it can be set at once. If all state is reset every frame, then there's no point of state block, just set state directly. That would be like creating VAO in OpenGL every frame & setting all vertex attribute pointers on it on every frame.

If something goes black when integrating this in app/game that means not all state is set properly, and something is missing. That should be fixed instead. And not recreating state block every frame.

Also after this commit d3d9 demo crashes when window is resized. My original code does not crash.

Heep042:
Hmm, since this commit is mine, I will have to debug. I could change it up a bit, make the changes switchable with a compile time switch or something. Sorry for introducing regressions.

Heep042:
I just found this bug very very weird, as it was not easily reproducible across different systems and looked how ImGui handled the rendering, which is exactly the way it's done currently with the state recreation.

PR #559, opened 2017-11-25
Fix NK_INCLUDE_COMMAND_USERDATA usage with images.

### LINKED ISSUE #511: A suggestion for improving nk_draw_list_push_image when NK_INCLUDE_COMMAND_USERDATA is used
Since `userdata` applied in `nk_draw_list_push_command` its usage for images suffers from the following problems which both as I discovered located in `nk_draw_list_push_image` function:
* Since branch `if (prev->elem_count == 0)` applies only `texture` to `prev`, `userdata` is omitted for the first command.
* Due to condition `if (prev->texture.id != texture.id)` two commands with the same texture are merged, despite them having possibly different `userdata` which is most likely against the wishes of the user.

Predelnik:
If userdata is used with drawing images userdata was missing from
the first image command and also subsequent commands which draw the
same texture but with different user data were incorrectly merged together.

I believe that this shouldn't be the case if userdata is used for something like "custom shaders" as it is stated in the comment because the same image with different shaders should most likely be drawn with different commands.

Sadly formatting of these `#ifdef`s looks a bit nasty, especially one in the if condition. I'm open to any suggestions/fixes.

This fixes #511

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -7968,9 +7968,16 @@ nk_draw_list_push_image(struct nk_draw_list *list, nk_handle texture)
         nk_draw_list_push_command(list, nk_null_rect, texture);
     } else {
         struct nk_draw_command *prev = nk_draw_list_command_last(list);
-        if (prev->elem_count == 0)
+        if (prev->elem_count == 0) {
             prev->texture = texture;
-        else if (prev->texture.id != texture.id)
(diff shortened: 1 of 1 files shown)


vurtun:
Hey @Predelnik.
Thanks for the PR. I changed the formatting a little bit but yeah does not look like there is a lot that can be done to clean it up, but that should be fine.

PR #574, opened 2017-12-08
nk_layout_peek() bugfix

NuklearBomb:
Fixes boundary fetching issues present in widgets organized by row templates.

It's difficult to describe the bug that occurs without the fix, so here's a comparison video that shows the bug and the bugfix side-by-side:
https://www.youtube.com/watch?v=oKRfy61Q4wU


vurtun:
Thanks for the PR,
quite an interesting bug.

PR #578, opened 2017-12-14
Remove 30 lines of code

Mark-Weston:
I've spent a couple hours trying to make the codebase a little bit smaller. Here's what I've succeeded to do without sacrificing any functionality (I think). If you don't like something please let me know.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -4480,7 +4480,7 @@ nk_rectiv(const int *r)
 NK_API struct nk_vec2
 nk_rect_pos(struct nk_rect r)
 {
-    return nk_vec2(r.x, r.y):
+    return nk_vec2(r.x, r.y);
 }
 
 NK_API struct nk_vec2


nuklear.h
```diff
@@ -294,27 +294,19 @@ extern "C" {
 #define NK_STRING_JOIN_DELAY(arg1, arg2) NK_STRING_JOIN_IMMEDIATE(arg1, arg2)
 #define NK_STRING_JOIN(arg1, arg2) NK_STRING_JOIN_DELAY(arg1, arg2)
 
-#ifdef _MSC_VER
-#define NK_UNIQUE_NAME(name) NK_STRING_JOIN(name,__COUNTER__)
```
dumblob: I'm pretty sure, these macros and defines shall stay and not be removed.

nuklear.h
```diff
@@ -294,27 +294,19 @@ extern "C" {
...
 #define NK_UNIQUE_NAME(name) NK_STRING_JOIN(name,__LINE__)
-#endif
 
 #ifndef NK_STATIC_ASSERT
 #define NK_STATIC_ASSERT(exp) typedef char NK_UNIQUE_NAME(_dummy_array)[(exp)?1:-1]
 #endif
 
 #ifndef NK_FILE_LINE
-#ifdef _MSC_VER
-#define NK_FILE_LINE __FILE__ ":" NK_MACRO_STRINGIFY(__COUNTER__)
```
dumblob: Same as above.

nuklear.h
```diff
@@ -4185,7 +4177,7 @@ struct nk_context {
 #define NK_MAX_FLOAT_PRECISION 2
 
 #define NK_UNUSED(x) ((void)(x))
-#define NK_SATURATE(x) (NK_MAX(0, NK_MIN(1.0f, x)))
+#define NK_SATURATE(x) (NK_CLAMP(0, (x), 1.0f))
```
dumblob: Good catch!

nuklear.h
```diff
@@ -4339,19 +4331,6 @@ template<typename T> struct nk_alignof{struct Big {T x; char c;}; enum {
...
- * are set incorrectly by the compiler, and compile out if not */
-NK_STATIC_ASSERT(sizeof(nk_size) >= sizeof(void*));
-NK_STATIC_ASSERT(sizeof(nk_ptr) == sizeof(void*));
-NK_STATIC_ASSERT(sizeof(nk_flags) >= 4);
-NK_STATIC_ASSERT(sizeof(nk_rune) >= 4);
-NK_STATIC_ASSERT(sizeof(nk_ushort) == 2);
-NK_STATIC_ASSERT(sizeof(nk_short) == 2);
-NK_STATIC_ASSERT(sizeof(nk_uint) == 4);
-NK_STATIC_ASSERT(sizeof(nk_int) == 4);
-NK_STATIC_ASSERT(sizeof(nk_byte) == 1);
```
dumblob: Why to remove this? It's a very important part of the library. We shall **not** remove this at any cost.

nuklear.h
```diff
@@ -4475,10 +4454,8 @@ NK_API struct nk_rect
 nk_recti(int x, int y, int w, int h)
 {
     struct nk_rect r;
-    r.x = (float)x;
-    r.y = (float)y;
-    r.w = (float)w;
-    r.h = (float)h;
+    r.x = (float)x; r.y = (float)y;
+    r.w = (float)w; r.h = (float)h;
```
dumblob: This is IMHO not necessary, but if you like it.

nuklear.h
```diff
@@ -4503,30 +4480,13 @@ nk_rectiv(const int *r)
 NK_API struct nk_vec2
 nk_rect_pos(struct nk_rect r)
 {
-    struct nk_vec2 ret;
-    ret.x = r.x; ret.y = r.y;
-    return ret;
+    return nk_vec2(r.x, r.y);
```
dumblob: This will be slower if the compiler can't inline `nk_vec2()`.

nuklear.h
```diff
@@ -4503,30 +4480,13 @@ nk_rectiv(const int *r)
...
-{
-    struct nk_rect res;
-    r.w = NK_MAX(r.w, 2 * amount);
-    r.h = NK_MAX(r.h, 2 * amount);
-    res.x = r.x + amount;
-    res.y = r.y + amount;
-    res.w = r.w - 2 * amount;
-    res.h = r.h - 2 * amount;
-    return res;
+    return nk_vec2(r.w, r.h);
```
dumblob: Same as above.

nuklear.h
```diff
@@ -4540,6 +4500,11 @@ nk_pad_rect(struct nk_rect r, struct nk_vec2 pad)
     return r;
 }
 
+NK_INTERN struct nk_rect nk_shrink_rect(struct nk_rect r, float amount)
+{
+    return nk_pad_rect(r, nk_vec2(amount, amount));
```
dumblob: Ok, this makes sense to me. Good catch!

nuklear.h
Mark-Weston: Why the special treatment for MSVC? gcc and clang have `__COUNTER__` as well, but `__LINE__` is good enough for them? Wouldn't `__LINE__` work just as well for MSVC too? I may be missing something since I've never used MSVC.

nuklear.h
Mark-Weston: Because it's redundant? It's defined in the header.

nuklear.h
Mark-Weston: This is to make it consistent with `nk_rect`, the function right above it.

nuklear.h
vurtun: It is for cases when `nuklear.h` is included two times once as header and later again inside a source file for implementation. In these cases both `include`s require the same defines. So to make sure that both times we have correct type sizes I added to compile time asserts once for the header and once for the implementation.

nuklear.h
vurtun: If it ain't broke don't fix it. Honestly I don't remember the reason. But I know that I got it from someone and he had a reason for it. However there is no reason to just delete something. Especially if you as well don't know the reason. Changing or removing something is always dangerous.

nuklear.h
vurtun: Yeah I don't really mind this change. It is more cosmetic than anything. Consistency is nice but on its on would not warrant a new version.

nuklear.h
Mark-Weston: @vurtun Agreed.

nuklear.h
Mark-Weston: @vurtun Correct me if I'm wrong, but if `nuklear.h` is included at least one, anywhere in the translation unit, be it directly in the source or indirectly through the `#include` chain, the static assertions are going to be there because they're in the API header part, and not the implementation part. To put it in other words, you can include the header without the implementation, but not the other way around (unless you define `NK_NUKLEAR_H_` before the `#include`, but then it won't even compile because of missing definitions).

So I just can't see any way to include the implementation without the API, but maybe I'm missing something.

PR #590, opened 2017-12-31
TravisCI changes

mgerhardy:
- always include the demo windows in compilation to check for errors
- compile the example
- use -Wall

### DIFF
--- a/demo/glfw_opengl3/Makefile
+++ b/demo/glfw_opengl3/Makefile
@@ -2,7 +2,7 @@
 BIN = demo
 
 # Flags
-CFLAGS = -std=c99 -pedantic -O2
+CFLAGS += -std=c99 -pedantic -O2
 
 SRC = main.c
 OBJ = $(SRC:.c=.o)
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,6 @@ before_install:
     - if [ $TRAVIS_OS_NAME == linux ]; then sudo add-apt-repository -y ppa:pyglfw/pyglfw && sudo apt-get update -qq && sudo apt-get install -y --no-install-recommends libglfw3 libglfw3-dev libglew-dev; fi
 
 script:
-    - make -C demo/glfw_opengl3
+    - make -C demo/glfw_opengl3 CFLAGS="-Wall -DINCLUDE_ALL"
     - make -C demo/glfw_opengl2
+    - make -C example
--- a/demo/allegro5/Makefile
+++ b/demo/allegro5/Makefile
@@ -2,7 +2,7 @@
 BIN = demo
 
 # Flags
-CFLAGS = -std=c99 -pedantic -O2
+CFLAGS += -std=c99 -pedantic -O2
 
 SRC = main.c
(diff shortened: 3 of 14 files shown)


mgerhardy:
See warnings shown in https://travis-ci.org/vurtun/nuklear/jobs/323619677

mgerhardy:
- What do you think about using -Werror for TravisCI? In my opinion a header only library should be free from warnings. So a CI should check for them.
- What do you think about activating appveyor? I could provide a CI config for it.

mgerhardy:
Rebased against latest master.

demo/glfw_opengl3/Makefile
```diff
@@ -2,7 +2,7 @@
 BIN = demo
 
 # Flags
-CFLAGS = -std=c99 -pedantic -O2
+CFLAGS ?= -std=c99 -pedantic -O2
```
dumblob: Well, in this case I would probably prefer `CFLAGS += -std=c99 -pedantic -O2`. What do you think?

mgerhardy: Changed. Also all other makefiles are changed now.

vurtun:
>- always include the demo windows in compilation to check for errors

Sure why not.

>- compile the example

Sure.

>- use -Wall

I compile with:

```
# Modes
.PHONY: clang
clang: CFLAGS += -Weverything -Wno-padded -Wno-missing-noreturn -Wno-covered-switch-default -Wno-float-equal -Wno-comma -Wthread-safety
clang: CFLAGS += -Wno-double-promotion -Wno-switch-enum -Wno-format-nonliteral
clang: CC = clang
clang: $(BIN)

.PHONY: gcc
gcc: CFLAGS += -Wall -Wextra -Wformat-security -Wunreachable-code -Wlogical-op
gcc: CFLAGS += -Winline -Wshadow -Wwrite-strings -Wdeprecated-declarations
gcc: CFLAGS += -Wstrict-prototypes -Wold-style-definition -Wconversion
gcc: CFLAGS += -Wredundant-decls -Wnested-externs -Wmissing-include-dirs
gcc: CFLAGS += -Wshadow -Wpointer-arith -Wcast-qual -Wcast-align -Wmissing-prototypes
gcc: CFLAGS += -Wswitch-default -Wundef -Wno-unused -Wstrict-overflow=5 -Wsign-conversion
gcc: CFLAGS += -Winit-self -Wstrict-aliasing -Wconversion -Wshift-count-negative -Wshift-count-overflow
gcc: CFLAGS += -Wswitch-enum -Winvalid-pch -Wbad-function-cast
gcc: CFLAGS += -Wno-switch-enum
gcc: CC = gcc
gcc: $(BIN)

```
I don't want these flags inside the provided Makefile. However if you are talking about CTravisCI than yeah no  problem having -Wall specified.

>- What do you think about using -Werror for TravisCI? In my opinion a header only library should be free from warnings. So a CI should check for them.

-Werror is a bad idea. Problem are different compiler version generating different warnings.

>- What do you think about activating appveyor? I could provide a CI config for it.

Never worked with it but sure why not.

Issue #584, opened 2017-12-26
Compiler warnings gcc 7.2


mgerhardy:
```
/home/mgerhardy/dev/engine/src/modules/ui2/private/nuklear.h: In function ‘nk_hash nk_murmur_hash(const void*, int, nk_hash)’:
/home/mgerhardy/dev/engine/src/modules/ui2/private/nuklear.h:5596:16: warning: this statement may fall through [-Wimplicit-fallthrough=]
     case 3: k1 ^= (nk_uint)(tail[2] << 16);
             ~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/mgerhardy/dev/engine/src/modules/ui2/private/nuklear.h:5597:5: note: here
     case 2: k1 ^= (nk_uint)(tail[1] << 8u);
     ^~~~
/home/mgerhardy/dev/engine/src/modules/ui2/private/nuklear.h:5597:16: warning: this statement may fall through [-Wimplicit-fallthrough=]
     case 2: k1 ^= (nk_uint)(tail[1] << 8u);
             ~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/mgerhardy/dev/engine/src/modules/ui2/private/nuklear.h:5598:5: note: here
     case 1: k1 ^= tail[0];
     ^~~~

```
And one in the demo/overview.c

```
In file included from /home/mgerhardy/dev/engine/src/tests/testnuklear/TestNuklear.cpp:4:0:
/home/mgerhardy/dev/engine/src/tests/testnuklear/overview.c: In function ‘int overview(nk_context*)’:
/home/mgerhardy/dev/engine/src/tests/testnuklear/overview.c:894:32: warning: unused variable ‘item_padding’ [-Wunused-variable]
                 struct nk_vec2 item_padding;
                                ^~~~~~~~~~~~

```

mgerhardy:
https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

PR #591, opened 2017-12-31
Fixed parts of issue #584

mgerhardy:


vurtun:
Not really essential but still nice thanks.

PR #609, opened 2018-01-22
Add opengl 4 demo that uses bindless textures and persistent mapped vertex buffer

lumonix:
- Full disclosure: I compiled/tested this with GLAD not GLEW. So might want to do a quick test.

- Wasn't sure if you want all examples to use GLEW or if I should add my GLAD h/c into the sample.
   Let me know if you want a GLAD example and I can modify the PR to use that instead.

- Tested only on linux (Ubuntu 17.10) and Windows 10

The key difference to understand is that the texture.id stored by nuklear refers to an 'index' into an array that holds all bindless textures.

The vertex and element buffers are persistent mapped. I think that might improve performance over the gl3 version, but have not done any comparison.

### DIFF
--- a/demo/glfw_opengl4/Makefile
+++ b/demo/glfw_opengl4/Makefile
@@ -1,0 +1,26 @@
+# Install
+BIN = demo
+
+# Flags
+CFLAGS += -std=c99 -pedantic -O2
+
+SRC = main.c
+OBJ = $(SRC:.c=.o)
+
+ifeq ($(OS),Windows_NT)
+BIN := $(BIN).exe
+LIBS = -lglfw3 -lopengl32 -lm -lGLU32 -lGLEW32
+else
+	UNAME_S := $(shell uname -s)
+	GLFW3 := $(shell pkg-config --libs glfw3)
+	ifeq ($(UNAME_S),Darwin)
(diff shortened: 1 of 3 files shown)


lumonix:
As a side note:
- Instead of storing the tex ids and handles into an array inside gl4 backend we could also make room to store a 64bit 'tex handle' on the nk image. I opted for the array method since there is something elegant about just having an 'int id' for the texture on the nuklear side. But open to discussion what you think.

vurtun:
Hey @lumonix,
sorry I am currently still not able to run it. While mesa already has bindless texture it seems like archlinux still has not updated to the newest version. As soon as the new version is available I will check it out.

lumonix:
@vurtun sounds good. No rush on my end.

lumonix:
Follow up note.
The lines 406 and 407: 

   /* Lock buffer until GPU has finished draw command */
    nk_glfw3_lock_buffer();

They should be placed at the very end of 'nk_glfw3_render'.
You might get artifacts rendering if they are not moved.

Cewein:
@lumonix is there a way of doing it with sample2D, check the issue #794 there is some hardware incompatiblie with some graphic card and/or CPU.
been tested on a 7th gen i5, windows 10 and GTX 1050 laptop, and a 8th gen i5 windows 10 and AMD RADEON 550 laptop same error : **'GL_ARB_gpu_shader_int64' not supported**

lumonix:
(I replied inside issue #794)

PR #646, opened 2018-03-18
Update codebase kLOC

KrzysztofSzewczyk:


vurtun:
Hey @KrzysztofSzewczyk,
at least [cloc](http://cloc.sourceforge.net/) calculates 18199 which is still in the 18kLOC range.

PR #655, opened 2018-03-30
Make text edit scroll less when typing past the end

mloccy:


a-roy:
We made this change because the current text edit behavior is unclear for users, particularly when pasting in long text. The current behavior of the text edit when a user types or pastes text past the end of the box makes it scroll past the end of the text, making it appear as though there is no text. This change makes it so it will only scroll by 0.25 of its width instead.

PR #661, opened 2018-04-05
Adding support for newlines in strings to nk_text/nk_label and family.

### LINKED ISSUE #532: Multiline text in labels
In nk_label or nk_label_wrap, when i displaying some text with symbols '\n' or '\r', all text showing in one line with symbols '?' instead of line wraping. What i'm doing wrong?

P.S. Sorry for my English, by the way.

AntiBlueQuirk:
The new behavior is as follows:
- nk_text/nk_label:
	One line of text will be rendered, up to the first newline
- nk_text_wrap/nk_label_wrap:
	Multiple lines of text will be rendered, wrapping when horizontal space
	is exhausted (as normal), but also when a newline is reached.
- nk_text_multiline/nk_label_multiline (new):
	Multiple lines of text will be rendered, wrapping only when a newline is reached.

This PR adds va_list functions, so it includes my pr_valist pull request (#660). If you don't want to merge pr_valist, I can provide a version without the va_list functions.

This PR should fix #532, though it does introduce a slight breaking change to nk_text and nk_text_wrap: Before those functions would render newline characters as ? symbols, and ignore them. In particular, nk_text would take entire paragraphs of text and render them on one line. Now it only renders the first line of text, and nk_text_wrap manually breaks on newlines. Since this behavior was probably undesirable, I doubt anyone is depending on it. The external behavior of these functions doesn't change, just they way they render things.


AntiBlueQuirk:
I'd like to apologize for the merge errors. I originally made my changes right before you did the header split, so when I went to merge the changes back in, I had to do so manually.

vurtun:
Hey @AntiBlueQuirk,
thanks for your PR I tried it out and had a number of issues. Many are small and I will probably fix them before pulling. However there are a few big issues. 

fuck_gimp

Using `nk_label_multiline` will cut of text (first line ends at "h" instead of displaying the same text as in the "Wrapped Text" section. The reason is that only 200 pixel of space in horizontal direction is provided but not handled correctly in the text widget.

Code used to test your implementation.
```c
nk_layout_row_static(ctx, 100, 200, 1);
nk_label_multiline(ctx, "This is a very long line to hopefully get this text to be wrapped into multiple lines to show line wrapping");
nk_layout_row_dynamic(ctx, 100, 1);
nk_label_multiline(ctx, "This is another long text to show dynamic window changes on multiline text");

nk_layout_row_static(ctx, 100, 200, 1);
nk_label_wrap(ctx, "This is a very long line\n to hopefully get this text to be wrapped into multiple lines\n to show line wrapping");
nk_layout_row_dynamic(ctx, 100, 1);
nk_label_wrap(ctx, "This is another long text to show dynamic window\n changes on multiline text");
nk_tree_pop(ctx);
```

Another problem is that your new implementation does wrapping on character basis and not word basis like previously implemented in nuklear (master):

word_wrap

nuklear (PR):

no_word_wrap

AntiBlueQuirk:
The multiline functions cutting off text is intentional. I added these new variants so that you can have text that only has manual line breaks.

I didn't realize I broke word wrapping. I tried to be careful not to do this. I guess I'll have to look at it some more.

vurtun:
> The multiline functions cutting off text is intentional. I added these new variants so that you can have text that only has manual line breaks.

Ah OK. Could you add some example code to `overview.c` to showcase what the use cases are for your each `nk_label_multiline` and `nk_label_wrap`. If I cannot make out what each of them can and cannot do then normal users won't know either.

> I didn't realize I broke word wrapping. I tried to be careful not to do this. I guess I'll have to look at it some more.

It is relatively simplistic word wrapping using spaces so nothing too complex.

AntiBlueQuirk:
Okay, It turns out I just inverted the wrapping logic, so easy fix. I also added some code to the demo to show the difference in behavior. Although, in retrospect, the naming the new functions `_multiline` might not be very clear. Perhaps it would better to call them something else, like `_manual`?

dumblob:
Speaking about word wrapping, don't forget about very long words, which won't fit in the text/label widget width. These exceptions should be character-wrapped instead of word-wrapped, otherwise (significant) part of the long word won't show up at all.

AntiBlueQuirk:
@dumblob I only changed the breaking functions just enough to add the new behavior, so the old behavior should still be there. I checked, and it looks like very short lines still break words by character when necessary.

tuket:
Is it possible to use this function yet?

PR #701, opened 2018-06-21
overview.c: Fix an out of bound access

buserror:
Access 8 elements in a 4 elements array.

Signed-off-by: Michel Pollet <buserror@gmail.com>


buserror:
Was just some cursory reading and this one jumped at me...

PR #726, opened 2018-08-19
fix unaligned access in murmur hash

grynca:
On asm.js and ARM 32bit int access must be aligned, otherwise it throws and exception.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -7153,23 +7153,23 @@ nk_murmur_hash(const void * key, int len, nk_hash seed)
 {
     /* 32-Bit MurmurHash3: https://code.google.com/p/smhasher/wiki/MurmurHash3*/
     #define NK_ROTL(x,r) ((x) << (r) | ((x) >> (32 - r)))
-    union {const nk_uint *i; const nk_byte *b;} conv = {0};
-    const nk_byte *data = (const nk_byte*)key;
-    const int nblocks = len/4;
+
     nk_uint h1 = seed;
+    nk_uint k1;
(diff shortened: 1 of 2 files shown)


nuklear.h
```diff
@@ -7153,23 +7153,23 @@ nk_murmur_hash(const void * key, int len, nk_hash seed)
...
     int i;
 
     /* body */
     if (!key) return 0;
-    conv.b = (data + nblocks*4);
-    blocks = (const nk_uint*)conv.i;
-    for (i = -nblocks; i; ++i) {
-        k1 = blocks[i];
+    for (i = 0; i < nblocks; ++i, keyptr += bsize) {
+        memcpy(&k1, keyptr, bsize);
```
dumblob: Did you do any performance measurements? Hashing is a bottle neck and if it should get slower, then I would discourage from using `memcpy` and rather implement a customized copy code (e.g. a plain `for` loop) perfectly fitting just this situation (instead of relying on external implementations which is in case of performance bottle necks quite dangerous).

Not speaking about the fact that up until now a totally independent single header library should get dependent as well as unpredictable regarding performance of hashing.

grynca: Well i did not run any measurements. I suppose i dont consider memcpy of 4 bytes to local variable as much of a dependency or bottleneck. Of course it could be changed to loop with 4x byte assignments but is it really neccessary?
The problem is that input data pointer is not 32bit aligned and so you cant cast it to 32bit int. Copying to local variable solves the issue.

dumblob:
@vurtun now it seems OK from my point of view.

PR #761, opened 2018-10-22
I removed an additional #include <math.h>

sukhchainn:
I found an additional "#include <math.h>" so I removed it.
Thinking that it was additional.


eax0r:
Bump, all good.

dumblob:
Done then :wink:.

PR #765, opened 2018-10-30
Fix issue 548

rswinkle:
The issue was introduced in the fix for 416 (commit 8559aeb).

The original problem was (backspace and arrows moving too fast) is not a
nuklear problem but a GLFW problem.  Because of the way nuklear handles
input (it must be between input_begin and end), the key callback method
GLFW offers is not an option.

So we have to use glfwGetKey() which returns the current state of a key
not whether it was just pressed or released, so nuklear acts like it
was pressed every single frame, hence the "too fast" problem.

The fix checks for state change and discards the event if there was
no change.  This kills key repeat behavior (for named keys) on
*all* platforms which makes deleting or arrowing inconvenient.

Since there's no way to fix the shortcomings of a callback vs event
based input API, my "fix" just makes the original fix conditional
on a macro NK_KEYSTATE_BASED_INPUT

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -106,6 +106,7 @@
 /// NK_BUTTON_TRIGGER_ON_RELEASE    | Different platforms require button clicks occurring either on buttons being pressed (up to down) or released (down to up). By default this library will react on buttons being pressed, but if you define this it will only trigger if a button is released.
 /// NK_ZERO_COMMAND_MEMORY          | Defining this will zero out memory for each drawing command added to a drawing queue (inside nk_command_buffer_push). Zeroing command memory is very useful for fast checking (using memcmp) if command buffers are equal and avoid drawing frames when nothing on screen has changed since previous frame.
 /// NK_UINT_DRAW_INDEX              | Defining this will set the size of vertex index elements when using NK_VERTEX_BUFFER_OUTPUT to 32bit instead of the default of 16bit
+/// NK_KEYSTATE_BASED_INPUT         | Define this if your backend uses key state for each frame rather than key press/release events
 ///
 /// !!! WARNING
 ///     The following flags will pull in the standard C library:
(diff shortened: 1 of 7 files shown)


nuklear.h
```diff
@@ -13922,8 +13922,12 @@ nk_input_key(struct nk_context *ctx, enum nk_keys key, int down)
     NK_ASSERT(ctx);
     if (!ctx) return;
     in = &ctx->input;
+#ifdef NK_KEYSTATE_BASED_INPUT
```
dumblob: Please write a short comment answering the questions "Why is it here? And why does it look like this and not otherwise?".

rswinkle: I explain in detail why it's there in the commit message and I think the name is pretty explanatory.  

If you're talking about an actual code comment right there explaining what it does, I think that's overkill and doesn't match the style of the code in nuklear.  There are several examples (ie line 7152, 8691, 8716 etc.) with #ifdef's in the code and no comments, and that's not even counting when entire functions are #ifdefed without comments.

I admit that I should have added a descriptive comment on line 109 where all the configuration macros are documented.  I should have thought to look for that.  I'll do that and resubmit, though it'd be easy enough for anyone with commit access to just add that since I selected the "maintainer can make changes" box.

dumblob: >a descriptive comment on line 109 where all the configuration macros are documented.

This is exactly what I meant, thanks. And sorry for not being clear from the beginning.

dumblob:
Now when it looks technically OK, could you please put the changes into corresponding files in `src/` and run `src/paq.sh` to generate ("amalgamate") `nuklear.h`?

Also don't forget updating the version in CHANGELOG :wink:.

rswinkle:
I'm such an idiot.  I can't believe I didn't even notice that and I should have know because any large single file library (or even 2 like SQLite) uses a similar system and I have one myself.

So I made the changes in the appropriate src files and regenerated but it looks like they were out of sync so this commit fixed many typos in and clarified portions of, documentation in the comments as well.

vurtun:
Thanks @rswinkle for your work and this PR.

dumblob:
>I'm such an idiot. I can't believe I didn't notice that and I should have known because any large single file library (or even 2 file like SQLite) uses a similar system and I have one myself.

I would argue you're not. It's normal and that's one of the reasons we have reviews :wink: Thanks!

PR #779, opened 2018-12-15
Fix typo in document.

xieyubo:


eax0r:
Fixed in #913, and this is wrong it should not be changed to "ratio", but "size".

- First example is **pixel values**. (width)
- Second example is **window ratio values**.
- Third example is **window ratio vales**.

@dumblob Remember to close this :smirk:.

dumblob:
Thanks for the effort. Closing as suggested by @eax0r .

PR #783, opened 2018-12-17
Add nk_***_get_scroll and nk_***_set_scroll for groups, windows, and popups

keharriso:
I thought these two functions (`nk_group_get_scroll` and `nk_group_set_scroll`) would be useful. I also added similar functions for windows and popups.

PR #802, opened 2019-02-18
Add nk_***_get_scroll and nk_***_set_scroll for groups, windows, and popups

keharriso:
Split #783 into two. This half of the changes adds `nk_***_get_scroll` and `nk_***_set_scroll` for groups, windows, and popups. It also bumps the version to 4.01.0

### DIFF
--- a/doc/nuklear.html
+++ b/doc/nuklear.html
@@ -2307,6 +2307,7 @@ X...XXXXXXXXXXXXX...X -           "
    - [x]: Major version with API and library breaking changes
    - [yy]: Minor version with non-breaking API and library changes
    - [zz]: Bug fix version with no direct changes to API
+- 2018/12/18 (4.01.0) - Added nk_***_get_scroll and nk_***_set_scroll for groups, windows, and popups
 - 2018/10/31 (4.00.2) - Added NK_KEYSTATE_BASED_INPUT to "fix" state based backends
 - 2018/04/01 (4.00.1) - Fixed calling `nk_convert` multiple time per single frame
 - 2018/04/01 (4.00.0) - BREAKING CHANGE: nk_draw_list_clear no longer tries to
--- a/doc/nuklear.html
+++ b/doc/nuklear.html
@@ -97,6 +97,7 @@ NK_INCLUDE_COMMAND_USERDATA     | Defining this adds a userdata pointer into eac
 NK_BUTTON_TRIGGER_ON_RELEASE    | Different platforms require button clicks occurring either on buttons being pressed (up to down) or released (down to up). By default this library will react on buttons being pressed, but if you define this it will only trigger if a button is released.
 NK_ZERO_COMMAND_MEMORY          | Defining this will zero out memory for each drawing command added to a drawing queue (inside nk_command_buffer_push). Zeroing command memory is very useful for fast checking (using memcmp) if command buffers are equal and avoid drawing frames when nothing on screen has changed since previous frame.
 NK_UINT_DRAW_INDEX              | Defining this will set the size of vertex index elements when using NK_VERTEX_BUFFER_OUTPUT to 32bit instead of the default of 16bit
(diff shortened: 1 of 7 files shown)


dumblob:
Thanks @keharriso, it looks good to me. It's clear, readable and simple. I didn't test it though.
@vurtun feel free to give it a short test and I think we can merge it.

irtusb:
what if I only want vertical scroll offset? nk_window_get_scroll() should accept NULL to ignore one or both args

keharriso:
> what if I only want vertical scroll offset? nk_window_get_scroll() should accept NULL to ignore one or both args. Same remark for the other two getters.

Done, thanks for the suggestion.

borodust:
Any chance this would be merged anytime soon?

dumblob:
Thanks for the reminder. Please rebase this pull on top of the current version and I'll merge it.

dumblob:
Thanks for rebasing - there are though some conflicts (even though GitHub doesn't show any conflicts for some reason) - see e.g. the content of https://github.com/vurtun/nuklear/blob/master/demo/x11_rawfb/Makefile and the diff in this pull.

Could you please check the pull for such cases? Thank you for understanding.

keharriso:
I'm afraid I don't see the difference.
https://github.com/vurtun/nuklear/blob/master/demo/x11_rawfb/Makefile
and
https://github.com/keharriso/nuklear/blob/get-set-scroll/demo/x11_rawfb/Makefile
look exactly the same to me.

borodust:
I wonder if upstream master was force pushed or @keharriso fork got corrupted in some way. I've rebased @keharriso's branch onto vurtun/master and after a bit of skipping commits i've got nice history back, or so it seems: https://github.com/vurtun/nuklear/compare/master...borodust:get-set-scroll

doc/nuklear.html
```diff
@@ -2258,6 +2307,9 @@
    - [x]: Major version with API and library breaking changes
    - [yy]: Minor version with non-breaking API and library changes
    - [zz]: Bug fix version with no direct changes to API
+- 2019/06/23 (4.01.0) - Added nk_***_get_scroll and nk_***_set_scroll for groups, windows, and popups
+- 2019/06/12 (4.00.3) - Fix panel background drawing bug
```
borodust: I also got these lines after rebasing. Are those correct?

dumblob:
@keharriso Travis has updated their distro and we didn't update the script so far (any pulls welcome - my bet is, that the ppa is not any more required as everything is in the default repositories).

>I'm afraid I don't see the difference.

There is no difference, but this shouldn't happen and I didn't check all the rest of the pull because if such thing happens, it must be cleared up first.

Actually I've checked the branch `get-set-scroll` from @borodust and can confirm, that this one is totally clear and is exactly what I would expect in the diffs. @keharriso any suggestions what would be your preferred way to proceed?

keharriso:
I'm pretty sure I just did the rebase wrong. I don't have much experience with that.

keharriso:
I just force pushed the branch from @borodust

doc/nuklear.html
keharriso: I believe they reflect the changes in CHANGELOG

borodust: Ah, that's a generated file, my bad.

nuklear.h
```diff
@@ -2609,6 +2643,8 @@ NK_API struct nk_rect nk_layout_space_rect_to_local(struct nk_context*, struct n
 /// nk_group_scrolled_offset_begin  | Start a new group with manual separated handling of scrollbar x- and y-offset
 /// nk_group_scrolled_begin         | Start a new group with manual scrollbar handling
 /// nk_group_scrolled_end           | Ends a group with manual scrollbar handling. Should only be called if nk_group_begin returned non-zero
+// nk_group_get_scroll              | Gets the scroll offset for the given group
```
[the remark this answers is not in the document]
keharriso: Alright, hopefully that commit fixes the slashes.

borodust:
Thanks!

PR #836, opened 2019-05-10
Editfix1

lundril:
This fixes two problems I found in nk_do_edit().
1) scroll position changes when de-activating and then re-activating edit widget
2) auto-scroll does not work when pressing "home" followed by "end".


demo/x11/Makefile
```diff
@@ -1,5 +1,6 @@
 # Install
 BIN = zahnrad
+FIX = fix
```
dumblob: I think to better distinguish the two binaries/apps, we should rather use `EDITFIX` instead of just `FIX` (same goes for `fix` -> `editfix`).

demo/x11/Makefile
```diff
@@ -11,3 +12,8 @@ $(BIN):
 	@mkdir -p bin
 	rm -f bin/$(BIN) $(OBJS)
 	$(CC) $(SRC) $(CFLAGS) -D_POSIX_C_SOURCE=200809L -o bin/$(BIN) -lX11 -lm
+
+$(FIX):
+	@mkdir -p bin
+	rm -f bin/$(FIX)
+	$(CC) editfix.c $(CFLAGS) -D_POSIX_C_SOURCE=200809L -o bin/$(FIX) -lX11 -lm
```
dumblob: To maintain regularity, could you please add `SRC = main.c` and `OBJ = $(SRC:.c=.o)` for `editfix.c`?

One more thing - when you're already at changing the Makefile, could you please add `.PHONY` sections as in the [x11_opengl2](https://github.com/vurtun/nuklear/blob/master/demo/x11_opengl2/Makefile ) demo?

lundril: Ok I tried now to do two things:
1) I added the .PHONY targets as in the x11_opengl2 demo
2) I renamed everything to "edittest" since this is really just for checking it out and does not fix anything.
2) I now use an edittest.c which might be included into main.c for the X11 demo like the other parts (overview.c etc).

The editfix1 branch I created contains three independent changes in the sense that you might apply them separately and each of them should leave you with a compileable and working source code.
One of these changes is just adding a test windows to be able to check out the fix easily; so this is not even needed at all.

Since I am new to github: 
Should I split these into three different pull requests ? I am not sure what's best for you ...

so long
  ingo

0x0203:
What is the status of this? Are changes/fixes still pending?

demo/x11/Makefile
dumblob: Sorry for the delay. No need to split the requests - we're trying to be less formal, if possible :wink:.

Because it's an update of `nuklear.h`, don't forget to make the changes (only) in `src/...`, bump version and then generate `nuklear.h` using `src/paq.sh` (see `src/Readme.md`).

Looking forward to merging it.

PR #847, opened 2019-05-31
Implementing missing x11_rawfb features colored text and multi-color-rect

0x0203:
This implements rendering of colored text for the x11_rawfb demo and an implementation of rendering multi-colored rectangles. The latter is used by the color picker widget which currently doesn't display properly. 

Both of these changes make use of the `nk_image_blendpixel()` routine which I modified to swap the `r` and `b` bytes of the nk_color. I'm still not sure where these are getting swizzled in the first place, but this change makes it so everything renders in the requested color. If anyone has any insight into where the 'real' fix for this is, I can make the appropriate change.

### DIFF
--- a/demo/x11_rawfb/nuklear_rawfb.h
+++ b/demo/x11_rawfb/nuklear_rawfb.h
@@ -70,7 +70,7 @@ struct rawfb_context {
 #endif
 
 static unsigned int
-nk_int_from_color(const struct nk_color *c)
+nk_rawfb_color2int(const struct nk_color *c)
 {
     unsigned int res = 0;
 #if defined(RAWFB_RGBX_8888)
@@ -90,7 +90,7 @@ nk_int_from_color(const struct nk_color *c)
 }
 
 static struct nk_color
-nk_color_from_int(const unsigned int i)
+nk_rawfb_int2color(const unsigned int i)
 {
     struct nk_color col;
 #if defined(RAWFB_RGBX_8888)
@@ -110,10 +110,10 @@ nk_color_from_int(const unsigned int i)
 }
 
 static void
-nk_rawfb_setpixel(const struct rawfb_context *rawfb,
+nk_rawfb_ctx_setpixel(const struct rawfb_context *rawfb,
     const short x0, const short y0, const struct nk_color col)
 {
-    unsigned int c = nk_int_from_color(&col);
+    unsigned int c = nk_rawfb_color2int(&col);
     unsigned char *pixels = rawfb->fb.pixels;
     unsigned int *ptr;
 
@@ -144,7 +144,7 @@ nk_rawfb_line_horizontal(const struct rawfb_context *rawfb,
 
     n = x1 - x0;
     for (i = 0; i < sizeof(c) / sizeof(c[0]); i++)
-        c[i] = nk_int_from_color(&col);
+        c[i] = nk_rawfb_color2int(&col);
 
     while (n > 16) {
         memcpy((void *)ptr, c, sizeof(c));
@@ -154,10 +154,10 @@ nk_rawfb_line_horizontal(const struct rawfb_context *rawfb,
 }
 
 static void
-nk_image_setpixel(const struct rawfb_image *img,
+nk_rawfb_img_setpixel(const struct rawfb_image *img,
     const int x0, const int y0, const struct nk_color col)
 {
-    unsigned int c = nk_int_from_color(&col);
+    unsigned int c = nk_rawfb_color2int(&col);
     unsigned char *ptr;
     unsigned int *pixel;
     NK_ASSERT(img);
@@ -175,7 +175,7 @@ nk_image_setpixel(const struct rawfb_image *img,
 }
 
 static struct nk_color
-nk_image_getpixel(const struct rawfb_image *img, const int x0, const int y0)
+nk_rawfb_img_getpixel(const struct rawfb_image *img, const int x0, const int y0)
 {
     struct nk_color col = {0, 0, 0, 0};
     unsigned char *ptr;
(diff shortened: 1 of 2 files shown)


demo/x11_rawfb/nuklear_rawfb.h
```diff
@@ -765,6 +846,10 @@ nk_rawfb_stretch_image(const struct rawfb_image *dst,
                     continue;
             }
             col = nk_image_getpixel(src, (int)xoff, (int) yoff);
+	    /* This assumes the font atlas uses transparent black for non-glyph pixels */
```
dumblob: Could you please also add this comment to the font baking code to make this assumption visible?

demo/x11_rawfb/nuklear_rawfb.h
```diff
@@ -171,10 +171,15 @@ nk_image_blendpixel(const struct rawfb_image *img,
     const int x0, const int y0, struct nk_color col)
 {
     struct nk_color col2;
+    struct nk_color t;
     unsigned char inv_a;
     if (col.a == 0)
         return;
 
+    t = col;
```
dumblob: This is really too hacky and I can't accept it, sorry.

demo/x11_rawfb/nuklear_rawfb.h
0x0203: I ended up simply removing this comment because ultimately, it doesn't matter. Either the pixel is 100% transparent, in which case it doesn't matter what color it is, or it's not, and it should be given the foreground color specified, leaving the alpha channel as given by the atlas.

demo/x11_rawfb/nuklear_rawfb.h
0x0203: No problem at all. Should now be properly fixed.

0x0203:
Okay, I think I've got this properly fixed now. As part of fixing this, I made some other changes that made things more consistent, clear, and correct. 
- The main fix was to make sure `getpixel()` and `setpixel()` read and write the pixel data in the same way.
- There were fencepost errors in the get/set pixel routines that prevented them from working on row/column 0. 
- Setting text color now applies all three color channels on any part of the font atlas image that is not 100% transparent.
- I made the names of the routines consistent to differentiate the functions that operate on a `struct rawfb_image` from those that deal with a `struct rawfb_context`.
- Fixes in main.c for the commented out `set_style()` function were made so that they would work with the rawfb context once uncommented. 
- Finally, to reduce confusion, I made the #define for the pixel ordering in main.c match that defined in the Makefile.

demo/x11_rawfb/nuklear_rawfb.h
```diff
@@ -70,26 +70,50 @@ struct rawfb_context {
 #endif
 
 static unsigned int
-nk_color_from_byte(const nk_byte *c)
+nk_int_from_color(const struct nk_color *c)
 {
     unsigned int res = 0;
 #if defined(RAWFB_RGBX_8888)
```
dumblob: Because `RAWFB_RGBX_8888` and `RAWFB_XRGB_8888` are mutually exclusive, could you extend all these ifdefs to `#if defined(RAWFB_RGBX_8888) && ! defined(RAWFB_XRGB_8888)` on all the places in the code?

Also I would prefer in this case pass by value as the dereferencing will be most probably more costly considering the size of the struct compared to the size of a pointer (because compilers are way more constrained when working with pointers than with copied values). Usually I wouldn't point this out, but this function is called for each pixel, which is definitely the critical path and such simple "optimization" is the lowest hanging fruit :wink:.

demo/x11_rawfb/nuklear_rawfb.h
```diff
@@ -130,19 +154,22 @@ nk_rawfb_line_horizontal(const struct rawfb_context *rawfb,
 }
 
 static void
-nk_rawfb_imagesetpixel(const struct rawfb_image *img,
+nk_image_setpixel(const struct rawfb_image *img,
```
dumblob: Could you please add some sane (short and unused by other backends) prefix to the function name (after `nk_`) and change also all other functions accordingly? I know you didn't introduce this mistake (it's our fault, that we approved previous pulls which introduced it), but when we're already by renaming...

The reason is to separate backends (but allow their parallel use - e.g. for runtime switching of backends) from Nuklear as backends are officially maintained by the community and not the Nuklear project (therefore backends are also not versioned and can change without a notice any time).

Second reason is to avoid confusion if Nuklear wanted to introduce similarly called function in the future.

0x0203: Makes sense. Left the bulk of the `nk_rawfb` prefixed functions alone and added `_ctx_` or `_img_` to the routines the appropriate get/set pixel functions that took in either a `struct rawfb_context` or `struct rawfb_image`. I also changed the to/from color function names because I have a thing against unreasonably long function names. Hopefully what I came up with is reasonable.

demo/x11_rawfb/nuklear_rawfb.h
0x0203: Done and done. While I'm content leaving the #define macros for RGBX/XRGB for now, the more I think about it, the more I think this shouldn't be a compile-time decision. This is all data that can be pulled from the FBIOGET_VSCREENINFO ioctl. If you don't mind, I'd like to get this PR closed out and I'll make a separate pull request after I get that designed and implemented properly. I'll probably add some things to the rawfb_image struct at init time to track pixel order and have the necessary routines check it on the fly.  I'll also need to check the other demos/back-ends to see if they need similar changes. But again... separate issue?

demo/x11_rawfb/nuklear_rawfb.h
0x0203: Thanks for the feedback; should now all be addressed.

0x0203:
Thanks for the feedback; should now all be addressed.

demo/x11_rawfb/nuklear_rawfb.h
dumblob: Currently the pull request is clean from my point of view and I would immediately merge it. But if you say you'll do a separate request, then you're welcome and I'm closing this one.

Thanks for your contributions!

dumblob:
Closing as per authors request - see https://github.com/vurtun/nuklear/pull/847#discussion_r290946425 33333 . There shall be another extended pull soon from the author.

PR #851, opened 2019-06-06
Implement missing features and make multiple fixes for x11_rawfb

0x0203:
This is effectively the same as PR #847 with a cleaned up commit history and the addition of making the pixel layout determined at runtime instead of relying on compile time decisions. This is done by storing the pixel layout (as an enum) in the `struct rawfb_image`. This is determined by reading the appropriate masks out of the XImage in `nk_xlib_init()` and setting the pixel layout appropriately. Currently only supports XRGB and RGBX. 

Also fixed up some potential null dereferencing issues on error return paths. 

The rest of the fixes are (see #847  for details):
- implementing multi color rectangle rendering
- implementing foreground text color rendering
- Cleanup and fixes of get/set pixel routines
- Fix arguments to `set_style()` examples.

### DIFF
--- a/demo/x11_rawfb/nuklear_rawfb.h
+++ b/demo/x11_rawfb/nuklear_rawfb.h
@@ -33,11 +33,18 @@
 
 struct rawfb_context;
 
+typedef enum rawfb_pixel_layout {
+    PIXEL_LAYOUT_XRGB_8888,
+    PIXEL_LAYOUT_RGBX_8888,
+}
+rawfb_pl;
+
+
 /* All functions are thread-safe */
-NK_API struct rawfb_context *nk_rawfb_init(void *fb, void *tex_mem, const unsigned int w, const unsigned int h, const unsigned int pitch);
+NK_API struct rawfb_context *nk_rawfb_init(void *fb, void *tex_mem, const unsigned int w, const unsigned int h, const unsigned int pitch, const rawfb_pl pl);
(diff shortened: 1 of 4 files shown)


demo/x11_rawfb/nuklear_rawfb.h
```diff
@@ -130,44 +173,49 @@ nk_rawfb_line_horizontal(const struct rawfb_context *rawfb,
...
+	ptr += img->pitch * y0;
+
         if (img->format == NK_FONT_ATLAS_ALPHA8) {
-            ptr += img->pitch * y0;
             col.a = ptr[x0];
             col.b = col.g = col.r = 0xff;
         } else {
-            ptr += img->pitch * y0;
-            col = ((struct nk_color *)ptr)[x0];
+	    pixel = ptr;
```
dumblob: Maybe directly `pixel = ((unsigned int *)ptr)[x0];`?

0x0203:
Good suggestions, thanks. I'll make another pull request. I'm still learning/getting used to the whole git/github workflow... should I create a new branch or if I continue to use this one for the new changes will everything still play nicely since your merge?

dumblob:
There shouldn't be any issues IMHO if you continue with this branch (judging based on the standard Git behavior). But I also don't have experience with exactly this case and GitHub, so let's try :wink:.

PR #856, opened 2019-06-12
Fix panel drawing bug where background wasn't drawn for last column

0x0203:
This is an issue where panels don't properly calculate the starting x position for filling the right empty space in `nk_panel_end()`. It results in much of the last column (the top empty space overlaps but has the correct width) of pixels not being drawn. Because the starting x position is based off of layout bounds instead of window bounds, it isn't necessary to subtract the border. 

This is most visible in renderers like x11_rawfb where there is no anti aliasing to blur the border and the background but occurs in all renderers. It effects anything that uses the panel, such as tooltips, combo boxes, menus, etc...

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -1514,7 +1514,7 @@ NK_API void nk_end(struct nk_context *ctx);
 /// Finds and returns a window from passed name
 ///
 /// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~c
-/// void nk_end(struct nk_context *ctx);
+/// struct nk_window *nk_window_find(struct nk_context *ctx, const char *name);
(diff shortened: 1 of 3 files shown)


0x0203:
Not sure why the CI tests are failing... looking at the logs, it doesn't seem to be related to this change, but because the build machine couldn't download the appropriate glfw libraries...?

dumblob:
Thanks for spotting the bg issue. Btw yes, it seems the newer Ubuntu is not supported by the ppa, so we'll need to take a look at it soon. Do you happen to know from which repositories to get all the 3 packages (`libglfw3 libglfw3-dev libglew-dev`)? Maybe in this Ubuntu (xenial) they're already there :man_shrugging:.

0x0203:
I'm not really sure... according to this: https://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=glfw
they should just be in there? I'm afraid I don't have much experience with Ubuntu packaging.

dumblob:
Well, yep, it seems all the three are there, so just removing the ppa addition shall fix it. If you have the time, issue a pull :wink:.

Issue #425, opened 2017-05-03
Support for subpixel correct drawing ?


LennertSchneider:
Hi,

I'm trying to render a waveform using nk_stroke_line, but it looks quite horrible, because all the float precision in the supplied coords is lost due to internal conversion to short.
I have the same problem with some other custom controls where I want to use lines and circles to construct them.
Is there a reason this has been done ? Is this really necessary to abandon the float precision internally ?

Cheers.

vurtun:
Before I start converting draw commands internal position type from short to float could you test something for me. `nk_convert_config` has parameters:
```c
struct nk_convert_config {
    /* ... */
    enum nk_anti_aliasing line_AA;
    unsigned circle_segment_count;
    unsigned arc_segment_count; 
    unsigned curve_segment_count;
    /* ... */
};
```
First could try turning line_AA anti-aliasing on and off. If that does not help please try to set higher values for each of the segment counts. These basically dictate out of how many triangles shapes/lines generated out of. Also for any curved shapes or lines I would highly recommend overriding `NK_SIN` and `NK_COS` to standard library functions `sin` and `cos`. 

If your problem still persist then I will change internal command types from `short` to `float`. As for the reason why they are `short`. This library started as an UI for X11 which only uses `short`. So at that time it made sense and took less memory. Especially since I at that did not know how much memory would, could and should be consumed.

LennertSchneider:
I think I might not have been clear with my statement. The problem is not in using the predefined circle/arc/curve-primitives, but with doing my own out of multiple lines. I first stumbled across this problem when trying to draw an arc without the begin/and strokes from the center (really JUST the arc). Since there was no primitive for this I tried to draw a bunch of lines whose coords I calc myself (using the precise stdlib sin/cos). However the resulting arc looks horrorbly deformed due to each start and endpoint of each line segment snapping to fill pixels, because of the internal short conversion.

itsuhane:
Sorry to bring this issue up again...I'm also facing a similar situation where I would like to have some custom widget displaying a function plot.

Below is a screenshot of my attempt drawing `sin(x)`.
![image](https://user-images.githubusercontent.com/1052233/59488514-5604f880-8eb2-11e9-91e2-ee8b6c22c8f0.png)
Like @LennertSchneider did, I sampled the function with different `x`, then plot the curve with `nk_stroke_polyline`.

Internally, Nuklear does float-to-short conversion:
[...]

PR #861, opened 2019-06-16
Add subpixel API.

itsuhane:
This PR tries to provide subpixel drawing API as discussed in #425.

In this PR, a new macro, `NK_ENABLE_SUBPIXEL_API` is proposed. User can turn-on subpixel API by defining this macro before **ALL** inclusion of `nuklear.h`.

When this macro is defined, the following changes are made inside nuklear:

- All draw command structs (except `nk_command_custom`) will use `float` instead of `short`/`unsigned short`, use `nk_vec2` instead off `nk_vec2i`.
- All draw command routines (except `nk_push_custom`) will have a `subpixel` counterpart. For example, in addition to `nk_stroke_polyline`, `nk_stroke_polyline_subpixel` will be defined. In these subpixel version, there will be no narrowing from `float` to `short`.

No documentation is changed currently.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -4540,6 +4540,29 @@ NK_API void nk_draw_text(struct nk_command_buffer*, struct nk_rect, const char *
 NK_API void nk_push_scissor(struct nk_command_buffer*, struct nk_rect);
 NK_API void nk_push_custom(struct nk_command_buffer*, struct nk_rect, nk_command_custom_callback, nk_handle usr);
 
+/* subpixel APIs */
+#ifdef NK_ENABLE_SUBPIXEL_API
+NK_API void nk_stroke_line_subpixel(struct nk_command_buffer *b, float x0, float y0, float x1, float y1, float line_thickness, struct nk_color);
+NK_API void nk_stroke_curve_subpixel(struct nk_command_buffer*, float, float, float, float, float, float, float, float, float line_thickness, struct nk_color);
+NK_API void nk_stroke_rect_subpixel(struct nk_command_buffer*, struct nk_rect, float rounding, float line_thickness, struct nk_color);
+NK_API void nk_stroke_circle_subpixel(struct nk_command_buffer*, struct nk_rect, float line_thickness, struct nk_color);
+NK_API void nk_stroke_arc_subpixel(struct nk_command_buffer*, float cx, float cy, float radius, float a_min, float a_max, float line_thickness, struct nk_color);
+NK_API void nk_stroke_triangle_subpixel(struct nk_command_buffer*, float, float, float, float, float, float, float line_thichness, struct nk_color);
+NK_API void nk_stroke_polyline_subpixel(struct nk_command_buffer*, float *points, int point_count, float line_thickness, struct nk_color col);
+NK_API void nk_stroke_polygon_subpixel(struct nk_command_buffer*, float*, int point_count, float line_thickness, struct nk_color);
+
+NK_API void nk_fill_rect_subpixel(struct nk_command_buffer*, struct nk_rect, float rounding, struct nk_color);
+NK_API void nk_fill_rect_multi_color_subpixel(struct nk_command_buffer*, struct nk_rect, struct nk_color left, struct nk_color top, struct nk_color right, struct nk_color bottom);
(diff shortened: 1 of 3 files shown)


dumblob:
Thanks @itsuhane . An important question - it seems to me, that now it behaves differently than you proposed in #425:

>As a proposal, there could be a global macro like `NK_VERTEX_VALUE_TYPE`, which defaults to `short`, but allow user to override with `float` to support subpixel drawing.
>Current draw routines will keep the casting to make sure they don't broke any pixel-perfect drawing for UI elements. Meanwhile, they will have subpixel counterparts, `nk_stroke_polyline_subpixel` for example, for any user who want advanced control on their drawing.

use case | `NK_ENABLE_SUBPIXEL_API` defined | `NK_ENABLE_SUBPIXEL_API` **NOT** defined
---|---|---
old codebase | all pixel-perfect calculations break (as `nk_stroke_polyline` behaves differently than originally) | no change, everything works
new codebase | works as expected (`nk_stroke_polyline` behaves the same as `nk_stroke_polyline_subpixel`) | the old Nuklear behavior (i.e. internal conversion `float` -> `short`)

Thinking about this further we might actually afford making `nk_stroke_polyline` (and others) behave differently (i.e. not convert `float` to `short` internally) if and only if `NK_ENABLE_SUBPIXEL_API` (or rather `NK_ENABLE_SUBPIXEL_PRECISION` or alike as it has nothing to do with API) is defined (by default this macro won't be defined). So actually there shouldn't be any need for `_subpixel()` counterparts and the above table would look like this:

use case | `NK_ENABLE_SUBPIXEL_API` defined | `NK_ENABLE_SUBPIXEL_API` **NOT** defined
---|---|---
old codebase | all pixel-perfect calculations break (as `nk_stroke_polyline` behaves differently than originally) | no change, everything works
new codebase | works as expected (`nk_stroke_polyline` doesn't do any `float` -> `short` conversions internally) | the old Nuklear behavior (i.e. internal conversion `float` -> `short`)

Btw please also bump version accordingly :wink:.

itsuhane:
Thanks for your quick response!

Since we changed the internal representation for `struct nk_command_*`, some pixel-perfect version must be changed accordingly. But their behavior was not changed: In the old codebase, we perform `float`-to-`short` conversions when creating these command structs. But before drawing, these `short` data are converted back to `float`, for example in line 1294 in `nk_convert`:

https://github.com/vurtun/nuklear/blob/bb327b97a98d865984263ce6a3d8c50739f8dcb1/src/nuklear_vertex.c#L1291-L1297

Effectively, the `float`-to-`short` conversion is just rounding-off the decimal parts. I think we can safely use `float` in commands without break the pixel-perfect drawing.

Originally, I was thinking about having a macro named `NK_VERTEX_VALUE_TYPE`, and user can define it to `float`. However, there are three types involved in the command structs: `nk_vec2i`, `short` and `unsigned short`, if the user chooses `float`, `unsigned float` is neither meaningful nor legal in C. And in some compilers, `unsigned float` will be interpreted as `unsigned int` without giving an error. That's why I choose a different macro.

If we give up the conversion, all UI rendering will no-longer be pixel-perfect. I believe it will introduce artifacts. If there are artifacts, then we cannot easily remove the conversion. Two separate set of drawing API must co-exist. I will try subpixel UI drawing first and make some screenshot.

But in some sense, I agree to have only one set of implementation because my current PR contains massive code duplication. Can I create some macro ("templated C") to help elide the duplication ?

dumblob:
>If we give up the conversion, all UI rendering will no-longer be pixel-perfect. I believe it will introduce artifacts.

This might be the case, but I think there is one problem which I'm reluctant to digest :wink:. Namely, that Nuklear users won't have any clue which API (one of pixel-perfect and subpixel) in which case to use as they won't know how it behaves and how to combine the APIs. We must not introduce such problem at any cost.

>I will try subpixel UI drawing first and make some screenshot.

This would be great - and once you're at it, try as many backends as you can to get some overview. This will be much appreciated.

>Can I create some macro ("templated C") to help elide the duplication ?

Sure, just make some proposals a and we'll discuss it. Usually one comes with the best ideas when working with code, but not up front, so I won't comment technically on this until there is some practical demonstrations :wink:.

PR #862, opened 2019-06-16
remove_extra_lines

### LINKED ISSUE #859: sfml_opengl3\main: extraneous lines break the program
this file:
nuklear-master\demo\sfml_opengl3\main.cpp
basically doesn't compile at the moment because there are 2 garbage lines in the source file.  If you delete these out, the program compiles and runs great.

The first is on line 174, the one reading:
`float bg[4];`
This line causes the version of "bg" on line 109 to be occluded since it's in a previous stack frame.  It also causes the line on 178 to have syntax errors, which ultimately kills the compile:
`glClearColor(bg.r, bg.g, bg.b, bg.a);`
The errors being of course, you can't use the period operator to access float array indexes.  Note, 178 is correct and needs to stay.

The second garbage line is the one on 176, because the second argument to the function there, "background", is undefined:
`nk_color_fv(bg, background);`
Looking at the code, this seems to be conceptually related to the other garbage line, but doesn't make sense because we've spent lines 152-155 loading the nk_colorf bg (from line 110) with colors, so why would we need to do it again?

Anyway, we should just delete 174 and 176 and call it a day.

crolando:
remove extraneous lines.  Fixes #859

### DIFF
--- a/demo/sfml_opengl2/main.cpp
+++ b/demo/sfml_opengl2/main.cpp
@@ -165,7 +165,6 @@ int main(void)
 
         /* Draw */
         win.setActive(true);
-        nk_color_fv(bg, background);
         glClear(GL_COLOR_BUFFER_BIT);
         glClearColor(bg.r, bg.g, bg.b, bg.a);
         /* IMPORTANT: `nk_sfml_render` modifies some global OpenGL state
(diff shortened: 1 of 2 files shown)


crolando:
Unfortunately this is completely redundant with #817, closing.

PR #867, opened 2019-06-20
Pointer size for MinGW 64Bit / Castings

Tropby:
The pointer size of the MinGW 64Bit pointer is "unsinged long long". 
Is this also true for GCC 64Bit on Linux?

There are a lot of compiler warnings at compile time for the "nuklear_gdi.h". 
I have added some castings.
For the GNUC compiler I added the pragmas to suppress warnings from newer C++ standards.

### DIFF
--- a/demo/gdi/nuklear_gdi.h
+++ b/demo/gdi/nuklear_gdi.h
@@ -13,6 +13,17 @@
 #ifndef NK_GDI_H_
 #define NK_GDI_H_
 
+// suppress compiler warnings for changes C89 -> C++11
+#ifdef __GNUC__
+    #if ( __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6) )
+        #pragma GCC diagnostic ignored "-Wpadded"
+        #pragma GCC diagnostic ignored "-Wold-style-cast"
+        #pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
+        #pragma GCC diagnostic ignored "-Wcast-align"
+        #pragma GCC diagnostic ignored "-Wcast-"
+    #endif
+#endif
+
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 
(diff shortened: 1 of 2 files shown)


demo/gdi/nuklear_gdi.h
```diff
@@ -62,28 +74,32 @@ nk_create_image(struct nk_image * image, const char * frame_buffer, const int wi
...
         image->region[1] = 0;
-        image->region[2] = width;
-        image->region[3] = height;
+        image->region[2] = (unsigned short)width;
+        image->region[3] = (unsigned short)height;
         
         INT row = ((width * 3 + 3) & ~3);
-        BITMAPINFO bi = { 0 };
+        BITMAPINFO bi;
+        memset( &bi, 0, sizeof(bi) );
```
dumblob: Is this required? It's less efficient and slightly "cumbersome" compared to the previous version.

dumblob (approved):
Looks really good to me, thanks. Just the one nitpick, but I'll leave it up to you to decide and afterwards I'll merge the pull :wink:.

demo/gdi/nuklear_gdi.h
Tropby: This is not required. 
The compiler got warnings because "0" is not equal "unsinged long".
If you changed the "0" to "0ul" the compiler said "0ul" is not equal to "int". 
I think the notation "{ 0 }" is not really designed for structs with different data types?

If you thinks its better to reverse this. Just do it.

dumblob: Hm, that's really weird. According to C99 and newer, the notation `{ 0 }` denotes the minimal case for partial struct assignment which in this case means assign `0` to the first struct member and all following struct members disregarding their type (which might be potentially harmful as e.g. `NULL` can be something else than `0` on some platforms or some IEEE 754 implementations could behave incorrectly if some flags are zero, but programmers for such platforms would know what to do with this source :wink:).

Are you using the provided `build.bat` script? Or some other compiler? What are all the options?

I would like to find out what's the issue in this case as it sounds really suspicious (might be also an issue with Microsoft Visual Studio compiler as used in `build.bat`).

PR #876, opened 2019-07-22
x11_xft fixes

Lory171:
Fixes for the X11 backend with XFT fonts from `demo/x11_xft`:
- fixed some memory leaks by destroying XFT objects during cleanup 
- the function to get text width was returning the wrong value
- the clipping space of text boxes was not implemented for XFT
- misc


dumblob:
Looks good to me, merging. Thanks!

PR #886, opened 2019-08-16
use NK_MEMSET instead of nk_memset in X11 demos

Hejsil:


dumblob:
Thanks for catching that!

PR #890, opened 2019-08-20
adds nk_tooltip_text

Hejsil:
For when you have a string that is not `\0` terminated

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -3524,6 +3524,7 @@ NK_API void nk_contextual_end(struct nk_context*);
  *
  * ============================================================================= */
 NK_API void nk_tooltip(struct nk_context*, const char*);
+NK_API void nk_tooltip_text(struct nk_context*, const char*, int);
 #ifdef NK_INCLUDE_STANDARD_VARARGS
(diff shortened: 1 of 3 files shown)


dumblob:
Hi, thanks for the proposal. To make this change happen, it would be wise to change (not add new API calls) the whole library to not use any null-terminated strings. For this reason I'm currently a bit reluctant to approve this PR. Would you be willing to make such an important step to happen?

Hejsil:
Hmm, idk if I'd want to put the time in make such a change. Though, I do mostly work with none null-terminated strings so I might end up needing to do this.

Would this involve removing APIs like `nk_label` and only having `nk_text`?

dumblob:
>Would this involve removing APIs like `nk_label` and only having `nk_text`?

In this case yes as the only motivation to have `nk_label` was to support null-terminated strings :wink: (back when Nuklear wasn't versioned and also when it became freshly versioned, it was needed to maintain backwards compatibility, but nowadays it's not such a big issue due to strict versioning).

Hejsil:
Closing this in favor of having this as an issue (#905)

PR #891, opened 2019-08-22
Added selected text bounds to `nk_tree_element_xxx`

Lory171:
For my project I needed to open a contextual menu when right clicking a tree node, but finding the widget bounds for the tree widget it's not possible with nuklear API and calculating them with precision from outside it's a mess.
While the `nk_tree_element_xxx` functions let you know if a tree node has been selected, it doesn't give you the bounds of the selection. I added exactly this, it may be useful for other situations like custom drawing on the node when selected.
I can write a different function just for this behaviour so to not change the current API, but I think it's right to have it in the selectable tree widget.
Feel free to close this pull request if you don't like the change.

### DIFF
--- a/demo/x11/nuklear_xlib.h
+++ b/demo/x11/nuklear_xlib.h
@@ -863,7 +863,7 @@ nk_xlib_shutdown(void)
     nk_xsurf_del(xlib.surf);
     nk_free(&xlib.ctx);
     XFreeCursor(xlib.dpy, xlib.cursor);
-    nk_memset(&xlib, 0, sizeof(xlib));
+    NK_MEMSET(&xlib, 0, sizeof(xlib));
 }
 
 NK_API void
--- a/demo/x11_rawfb/nuklear_rawfb.h
+++ b/demo/x11_rawfb/nuklear_rawfb.h
@@ -821,7 +821,7 @@ nk_rawfb_init(void *fb, void *tex_mem, const unsigned int w, const unsigned int 
     if (!rawfb)
         return NULL;
 
-    nk_memset(rawfb, 0, sizeof(struct rawfb_context));
+    NK_MEMSET(rawfb, 0, sizeof(struct rawfb_context));
     rawfb->font_tex.pixels = tex_mem;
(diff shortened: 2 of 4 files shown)


src/nuklear_tree.c
```diff
@@ -272,15 +272,15 @@ nk_tree_element_image_push_hashed_base(struct nk_context *ctx, enum nk_tree_type
     {nk_flags dummy = 0;
     struct nk_rect label;
     /* calculate size of the text and tooltip */
-    text_len = nk_strlen(title);
```
dumblob: Good catch, thanks. It seems `text_len` is not used anywhere then - could you also remove it (as well as the comment `/* calculate size of the text and tooltip */`)?

Lory171:
> Hm, did you try to use nk_popup_begin and friends first? I can't comment more without seeing your source code, but I would say there shall be no need for any bounds for use cases you've described.

The problem is not using `nk_contextual` or `nk_popup`, it's the fact that getting the area (position and size) of some widgets with precision it's not possibile, neither calculating it nor getting it with `nk_widget_bounds`, as you said the API it's not meant for this. The area is needed for both `nk_contextual` and `nk_popup`, as you need to know which area trigger the activation of the popup.
The problem with the tree widget it's that the allocation of space it's not controlled by the user but it happen inside the function call. Using `nk_widget_bounds` before gives you the area value of the last widget before the tree, using it after gives you the area in the next row. I think this can happen also with other complex widgets.

Lory171:
Calculating the width of text was messy so I added some easier functions for it
[Add nk_layout_get_xxx_width API functions](https://github.com/vurtun/nuklear/pull/891/commits/3efc0bbbe44211c8464f9c070f4c78c05b438e47)
[nk_layout_get_xxx_width fixes](https://github.com/vurtun/nuklear/pull/891/commits/de6e16843226efa3f9fb9af00e0a3720f5db3eb0)

Lory171:
I added more changes to nuklear, so I cleaned up and opened different pull requests for different topics.
Sorry for the mess.

PR #900, opened 2019-09-10
Fixing the nk_cos function.

molecularentropy:
Drawing a circle, with `config.circle_segment_count` set to `22`, one can see a small kink in the rightmost part of the circle. Increasing the segment count to 220, we see this:
Screenshot_2019-09-10_18-37-08
It seems that the nk_cos implementation is actually really off. I have regenerated the approximation series with `lolremez`. Now, with the fixed nk_cos implementation we get:
Screenshot_2019-09-10_18-48-24
Please consider for inclusion.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -6087,15 +6087,18 @@ nk_sin(float x)
 NK_LIB float
 nk_cos(float x)
 {
-    NK_STORAGE const float a0 = +1.00238601909309722f;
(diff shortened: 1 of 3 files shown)


dumblob:
Yep, `nk_cos` was always meant only for nuklear internal use for low number of pixels. But I see, that this small change which shall have rather small impact on performance makes a significant difference, so I'm definitely merging it :wink:.

Thanks for tracking this down and for the pull!

PR #902, opened 2019-09-13
Use C style comment

crazyBaboon:
C++ style comments are not allowed in ISO C89


dumblob:
Missed that one, thanks for catching!

PR #903, opened 2019-09-16
Compiler warning fix and two sanity changes

oj002:
I was annoyed by nuklear not cleanly building so I set out to fix that.
I also fixed some issuse I've spotted along the way.


nuklear.h
```diff
@@ -18545,7 +18545,6 @@ nk_tree_element_image_push_hashed_base(struct nk_context *ctx, enum nk_tree_type
     struct nk_vec2 item_spacing;
     struct nk_rect header = {0,0,0,0};
     struct nk_rect sym = {0,0,0,0};
-    struct nk_text text;
```
dumblob: Hi, thanks for the fixes. I totally forgot why we wanted to set/change background in this function - did you try to semantically analyze where it would fit?

Also please bump version in CHANGELOG, thanks :wink:.

dumblob (changes requested):

PR #880, opened 2019-08-04
- Fix a bug where combo box can't be closed by clicking the header wh…

kamil-zubair:
…en NK_BUTTON_TRIGGER_ON_RELEASE is defined

PR #908, opened 2019-09-20
Fix a bug wherein combobox cannot be closed by clicking the header + Update documentation link to use Statically (CLEAN)

### LINKED ISSUE #879: Wrong combo behavior when NK_BUTTON_TRIGGER_ON_RELEASE is defined
When NK_BUTTON_TRIGGER_ON_RELEASE  is defined, combobox is opened when mouse button is release as it should be.
But when I clicked on opened combo box, it closed on mouse click and opened again on mouse released.
After debugging the code I found that the culprit is in function nk_nonblock_begin in line 

`pressed = nk_input_is_mouse_pressed(&ctx->input, NK_BUTTON_LEFT);`

After I changed it to :
```
#ifdef NK_BUTTON_TRIGGER_ON_RELEASE
	pressed = nk_input_is_mouse_released(&ctx->input, NK_BUTTON_LEFT);
#else
        pressed = nk_input_is_mouse_pressed(&ctx->input, NK_BUTTON_LEFT);
#endif
```

the combo box works as intended.
Is this the right way to fix the issue ? Anyway, I hope the bug is fixed in the next release.

eax0r:
Hello,

Quick intro for @dumblob, I am a software engineer that really like nuklear (using it since early commits) and the concept of single-file library, I will try to be really active on nuklear and make it as clean as possible, and thus keep it always updated.
Thanks @vurtun for his amazing work and thanks @dumblob for still being such an active maintainer. 💯

Some people actually create new accounts, make a pull request then forget to make it compliant, and then the pull request become abandoned.
I am testing everything on multiple systems, backends, and on different scenarios, so If I feel that something had to be pushed upstream, I am going for it, many pull requests are doomed nowadays.

Here is the clean pull request (instead of #880, thanks @kamil-zubair for fixing the bug).
Then you can close #879.

P.S. can we fix Travis CI already... :smile:

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -25,7 +25,7 @@ render backends it only focuses on the actual UI.
 - No global or hidden state
 - Customizable library modules (you can compile and use only what you need)
 - Optional font baker and vertex buffer output
-- [Documentation](https://rawgit.com/vurtun/nuklear/master/doc/nuklear.html)
+- [Documentation](https://cdn.statically.io/gh/vurtun/nuklear/master/doc/nuklear.html)
 
 ## Building
 
--- a/nuklear.h
+++ b/nuklear.h
@@ -16928,7 +16928,11 @@ nk_nonblock_begin(struct nk_context *ctx,
     } else {
         /* close the popup if user pressed outside or in the header */
         int pressed, in_body, in_header;
+#ifdef NK_BUTTON_TRIGGER_ON_RELEASE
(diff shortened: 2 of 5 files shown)


dumblob:
Hi @eax0r that's amazing to hear. Just go ahead :wink:.

>P.S. can we fix Travis CI already... 

You know it already - make a PR and it's gone :smile:.

Issue #907, opened 2019-09-19
Internal drawing bug


eax0r:
Hello,

So after investigating a lot, I have found that the commit 9a9f4f063dcec097c34adbc2838c963602d0826e is actually breaking the **D3D9 backend**, and probably many other backends.

E.g. :  #480

I am working on many nuklear project for some time now, and expect some fixes from me.

I think we should revert the commit as this is a serious issue, and such a "ghetto way" to actually fix the real problem.

eax0r:
Note about the D3D9 backend:

For a **500 x 500 pixels** window:

When I revert the commit 9a9f4f063dcec097c34adbc2838c963602d0826e:

with **NK_ANTI_ALIASING_OFF**
The window is drawn at the correct size.
**result** in a 500px x 500px window.

with **NK_ANTI_ALIASING_ON**
The window is drawn with the 0.5f offsets (1px more).
**result** in a 501px * 501px window.

dumblob:
Hey, we're glad you're using Nuklear in many projects.

Regarding 1px off, feel free to make a pull request with the improvement. I'm actually quite confident we shouldn't just blindly revert the commit, but come up with a better solution - any ideas? Maybe add the "anti-aliasing" also to other parts of Nuklear?

eax0r:
Well, I have at least fixed a bug with `NK_ANTI_ALIASING_OFF`, and I can confirm it was indeed a regression introduced in the commit 9a9f4f063dcec097c34adbc2838c963602d0826e.

But I still have to figure out the whole anti-aliasing/clipping broken thing (nightmare with all the backends)...

eax0r:
Ok, I am implementing all the fixes for `NK_ANTI_ALIASING_OFF` right now, and then I will make a pull request if everything pass my tests.

When it will get merged to the upstream by @dumblob, I will then work on `NK_ANTI_ALIASING_ON`.

It looks good so far :wink:.

PR #910, opened 2019-09-21
Old commit.

eax0r:


eax0r:
This fix a bug when `NK_ANTI_ALIASING_OFF` is used, as reported here #907.

eax0r:
before
after

eax0r:
The fix is great, but do not merge yet, I am looking for other related things, and I will use native git and stop using the web interface resulting in multiple commits.

Thanks.

PR #912, opened 2019-09-22
Travis: Fix

eax0r:
@dumblob Travis CI should build ok now. It was not able to fetch the pyglfw repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,25 @@
 language: c
+sudo: required
+dist: xenial
 
 os:
-    - linux
+  - linux
 
 compiler:
-    - gcc
-    - clang
+  - gcc
+  - clang
 
 before_install:
-    - if [ $TRAVIS_OS_NAME == linux ]; then sudo add-apt-repository -y ppa:pyglfw/pyglfw && sudo apt-get update -qq && sudo apt-get install -y --no-install-recommends libglfw3 libglfw3-dev libglew-dev; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,16 +1,25 @@
...
-    - clang
+  - gcc
+  - clang
 
 before_install:
-    - if [ $TRAVIS_OS_NAME == linux ]; then sudo add-apt-repository -y ppa:pyglfw/pyglfw && sudo apt-get update -qq && sudo apt-get install -y --no-install-recommends libglfw3 libglfw3-dev libglew-dev; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then 
+      sudo apt-get update -qq;
+      sudo apt-get install -y --no-install-recommends libglfw3 libglfw3-dev libglew-dev;
+      wget https://github.com/glfw/glfw/releases/download/3.3/glfw-3.3.zip;
```
dumblob: Hm, I thought this shouldn't be necessary. Could you please shed some light on why do we need to compile our own glfw? Doesn't the package `libglfw3` and `libglfw3-dev` offer all the needed stuff? Compiling our own glfw feels suspicious :wink:.

eax0r: Indeed, thanks for catching that, I forgot that, used to download and build GLFW since sometimes APT repository are unreliable.

But you are right we do not need it here.

dumblob:
Done, thanks!

PR #918, opened 2019-10-01
icon_load() function not used

crazyBaboon:
This function was not used in example/canvas.c, so... commented out.


dumblob:
I actually think this might be useful to someone. It has already been written, so I'd just comment it to prevent unnecessary compilation and warnings, but I'd leave it there :wink:.

dumblob:
Thanks :wink:.

PR #923, opened 2019-10-09
Scrollfix1

lundril:
I now try to do this more slowly.
This is the first part of the previous pull request "Editfix1 #836".
I hope this is easier to merge/pull :-)


crazyBaboon:
Nuklear aims to be C89 compliant, so the comments should to be C style, not C++...

PR #930, opened 2019-10-25
Tidy up the README a little

sjg20:
Some of the sentences in the README are pretty long and some terms could
do with hyphens. Tidy this up.

Signed-off-by: Simon Glass <sjg@chromium.org>

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -2,25 +2,25 @@
 
 [![Build Status](https://travis-ci.org/vurtun/nuklear.svg)](https://travis-ci.org/vurtun/nuklear)
 
-This is a minimal state immediate mode graphical user interface toolkit
+This is a minimal-state, immediate-mode graphical user interface toolkit
 written in ANSI C and licensed under public domain. It was designed as a simple
 embeddable user interface for application and does not have any dependencies,
-a default render backend or OS window and input handling but instead provides a very modular
-library approach by using simple input state for input and draw
-commands describing primitive shapes as output. So instead of providing a
+a default render backend or OS window/input handling but instead provides a
+highly modular, library-based approach, with simple input state for input and
+draw commands describing primitive shapes as output. So instead of providing a
 layered library that tries to abstract over a number of platform and
-render backends it only focuses on the actual UI.
+render backends, it focuses only on the actual UI.
 
 ## Features
 
-- Immediate mode graphical user interface toolkit
-- Single header library
+- Immediate-mode graphical user interface toolkit
+- Single-header library
 - Written in C89 (ANSI C)
 - Small codebase (~18kLOC)
 - Focus on portability, efficiency and simplicity
 - No dependencies (not even the standard library if not wanted)
 - Fully skinnable and customizable
-- Low memory footprint with total memory control if needed or wanted
+- Low memory footprint with total control of memory usage if needed / wanted
 - UTF-8 support
 - No global or hidden state
 - Customizable library modules (you can compile and use only what you need)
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -93,7 +93,7 @@ nk_end(&ctx);
 
 ## Bindings
 There are a number of nuklear bindings for different languges created by other authors.
-I cannot atest for their quality since I am not necessarily proficient in either of these
+I cannot atest for their quality since I am not necessarily proficient in any of these
```
servant-42: @sglass68: Thanks! There's another typo here, "atest" should be "attest"

Readme.md
```diff
@@ -2,25 +2,25 @@
 
 [![Build Status](https://travis-ci.org/vurtun/nuklear.svg)](https://travis-ci.org/vurtun/nuklear)
 
-This is a minimal state immediate mode graphical user interface toolkit
+This is a minimal-state immediate-mode graphical user interface toolkit
```
servant-42: @sglass68: I don't think "minimal-state" is typically hyphenated (at least, in a little bit of googling I'm not finding examples of it that way)

Readme.md
```diff
@@ -2,25 +2,25 @@
...
 written in ANSI C and licensed under public domain. It was designed as a simple
-embeddable user interface for application and does not have any dependencies,
-a default render backend or OS window and input handling but instead provides a very modular
-library approach by using simple input state for input and draw
-commands describing primitive shapes as output. So instead of providing a
+embeddable user interface for applications. It does not have any dependencies,
```
servant-42: @sglass68: This is subjective, but IMO this is better as a single sentence due to the tight semantic connection between the two halves and also to have more [variety of sentence length](https://wordsbyevanporter.com/how-to-use-varied-sentence-lengths-for-better-writing/).

Readme.md
```diff
@@ -29,19 +29,19 @@ render backends it only focuses on the actual UI.
...
-The implementation mode requires to define  the preprocessor macro
+The implementation mode requires defining the preprocessor macro
 `NK_IMPLEMENTATION` in *one* .c/.cpp file before `#include`ing this file, e.g.:
 ```c
 #define NK_IMPLEMENTATION
 #include "nuklear.h"
 ```
 IMPORTANT: Every time you include "nuklear.h" you have to define the same optional flags.
-This is very important not doing it either leads to compiler errors or even worse stack corruptions.
+This is very important not doing it either leads to compiler errors, or even worse, stack corruptions.
```
servant-42: @sglass68: I think this also needs a semi-colon or colon to separate "This is very important" from the rest of the sentence.

prust (commented):
@sglass68: Thank you for fixing all these grammatical mistakes! I must not have looked very closely before because I didn't realize how many there were. I did leave some comments below with some suggestions, but some are subjective, so feel free to ignore if you disagree.

I'm sorry about the 2-week delay, @vurtun has been busy on other things for quite some time and @dumblob recently had some family issues come up, so this project hasn't had the maintenance attention that it deserves lately.

Readme.md
sjg20: I'm not sure google is going to help much. The way I read it, it is creating an adjective out of a noun (state).

What is this sentence intended to mean? I assumed:

It is a GUI toolkiit
It uses immediate mode
It keeps minimal state internally

'This is a minimal state' probably references to a country ;-)

Perhaps I should have suggested: This is a GUI toolkit which keeps minimal state and operates in immediate mode?

Readme.md
sjg20: yes

Readme.md
sjg20: yes

Readme.md
servant-42: @sglass68:

> it is creating an adjective out of a noun (state)

You're right, good point.

> Perhaps I should have suggested: ...

I'm happy either way (though I slightly prefer the `minimal-state immediate-mode` wording, as it is currently in this PR).

Issue #326, opened 2017-01-20
Scroll to bottom of a scrollbar


Wizzard033:
Is there a way to scroll to the bottom of a scrollbar? I'm using nk_group_scrolled_begin.
I tried setting the height to a ridiculously high number, but that makes it flash one frame.
Instead, I can render it twice every frame, but that seems sort of like a hack.

        nk.group_scrolled_begin(ctx, outputScroll, "Console Output", nk.WINDOW_BORDER)
        nk.layout_row_dynamic(ctx, 17, 1)
        for i = #outputLines, 1, -1 do
            nk.label_colored_wrap(ctx, outputLines[i][1], outputLines[i][2])
        end
        nk.group_scrolled_finalize(ctx)

I can do outputScroll[0].h = 1000000 then what I want happens, but the frame I do it it messes up.
Therefore, I have to run the above code 2x in one frame after setting the height to 1000000.

Wizzard033:
I tried calculating the height of the scrolled region in various ways, but I couldn't match the scrollbar.
Therefore, I have no idea what units the scrollbar is using.
Since I have no idea what units its using, I cannot effectively set it to be scrolled to the bottom.

DeXP:
`nk_group_scrolled_offset_begin(ctx, &scrollHorizOffset, &scrollVertOffset, winTitle, NK_WINDOW_BORDER)`
where `scrollHorizOffset` and `scrollVertOffset` are global or static variables. When you need to scroll down once - just change `scrollVertOffset = 1000000`.

Wizzard033:
As mentioned, that makes it flash one frame (implying it actually renders at 1000000 one time before correcting it's position). I was calling this very often, and so it was unacceptable for it to be flashing.

PR #931, opened 2019-10-26
Add nk_*_scroll_here to scroll to the current widget

keharriso:
Stops the flickering mentioned in #326 when scrolling beyond the end of a window, popup, or group.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -6087,8 +6087,8 @@ nk_sin(float x)
 NK_LIB float
 nk_cos(float x)
 {
-    /* New implementation. Also generated using lolremez. */
-    /* Old version significantly deviated from expected results. */
+    // New implementation. Also generated using lolremez.
+    // Old version significantly deviated from expected results.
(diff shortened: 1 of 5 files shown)


dumblob:
@eax0r I think you've spent a lot of time digging into these details - could you please review it considering your experience? Thanks a lot.

keharriso:
I think it's worth noting that I changed the way the functions work a bit. You need to call *_set_scroll some time after you reach the desired scroll width/height. So if you want to scroll to the end, you need to call *_set_scroll at the bottom of the body for a window, popup, or group.

keharriso:
I reverted *_set_scroll to the previous behavior and instead added *_scroll_here which scrolls to the current widget. If you call it at the end of a window/popup/group body, then it will scroll down to the end.

keharriso:
I can't reproduce the flickering issue, so I am closing this pull request. If the issue comes back, I'll be sure to make a new pull request.

